PR #3, opened 2025-10-14
Implement parquet backed storage

adsharma:
More context in this [blog post](https://adsharma.github.io/beating-the-CAP-theorem-for-graphs/)

```
# copy  https://github.com/adsharma/graph-std/tree/main/karate/karate_csr to ./karate_csr and then try

$ cat test.cypher
CREATE NODE TABLE Person(id INT64, club STRING, PRIMARY KEY(id)) WITH (storage = './karate_csr/karate_random');
CREATE REL TABLE knows(FROM Person TO Person, weight DOUBLE) WITH (storage = './karate_csr/karate_random');

echo "MATCH (u1)-[k]->(u2) RETURN u1.id, u1.club, k.weight, u2.id;"  | ./build/debug/tools/shell/lbug -i test.cypher
-- Processing: test.cypher
Opening the database under in-memory mode.
Enter ":help" for usage hints.
┌───────┬─────────┬──────────┬───────┐
│ u1.id │ u1.club │ k.weight │ u2.id │
│ INT64 │ STRING  │ DOUBLE   │ INT64 │
├───────┼─────────┼──────────┼───────┤
│ 333   │ Mr. Hi  │ 1.000000 │ 306   │
│ 333   │ Mr. Hi  │ 1.000000 │ 41    │
│ 333   │ Mr. Hi  │ 1.000000 │ 211   │
│ 333   │ Mr. Hi  │ 1.000000 │ 54    │
│ 333   │ Mr. Hi  │ 1.000000 │ 14    │
│ 302   │ Mr. Hi  │ 1.000000 │ 41    │
│ 268   │ Mr. Hi  │ 1.000000 │ 41    │
│ 306   │ Mr. Hi  │ 1.000000 │ 167   │
│ 75    │ Mr. Hi  │ 1.000000 │ 41    │
│ 75    │ Mr. Hi  │ 1.000000 │ 305   │
│  ·    │   ·     │    ·     │  ·    │
│  ·    │   ·     │    ·     │  ·    │
│  ·    │   ·     │    ·     │  ·    │
│ 167   │ Officer │ 1.000000 │ 41    │
│ 167   │ Officer │ 1.000000 │ 14    │
│ 87    │ Officer │ 1.000000 │ 268   │
│ 87    │ Officer │ 1.000000 │ 75    │
│ 87    │ Officer │ 1.000000 │ 54    │
│ 87    │ Officer │ 1.000000 │ 14    │
│ 87    │ Officer │ 1.000000 │ 161   │
[...]

PR #34, opened 2025-11-01
Use ccache

adsharma:
Keep the planet green.


adsharma:
ccache is effective. Build step completed in 30s instead of 25m in run #3 here vs run #2.

https://github.com/LadybugDB/ladybug/actions/runs/19002850417/job/54275375570

adsharma:
Test run for the effectiveness of `actions/cache` strategy.

https://github.com/LadybugDB/ladybug/actions/runs/19004368916

PR #38, opened 2025-11-02
Add AGENTS.md

adsharma:


adsharma:
Yes, this is for AI agents. Also serves as an acknowledgement of the new reality.

We want people to be responsible for the code they submit, but they shouldn't be shy about using better tools for the job.

PR #56, opened 2025-11-05
Unify nodejs builds

adsharma:
Also enable and test wasm builds


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/19090382818

PR #57, opened 2025-11-05
extensions: use lowercase in the URLs

adsharma:
Should fix the CI error we see


adsharma:
arm64 blocked by https://github.com/opt-nc/setup-duckdb-action/issues/218

PR #62, opened 2025-11-09
extensions: purge older releases

adsharma:
The current extension docker image is ~4GB
Older kuzu extensions probably won't work anyway. So purge them.


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/19217171400

PR #102, opened 2025-12-07
Support DuckDB as a foreign node table

adsharma:
projection/limit/filter push downs are working

```
attach 'test.duckdb' as wd (dbtype duckdb);
match (a: wd.nodes) return a.id, a.text, a.docid limit 5;
match (a: wd.nodes) where a.id < 10 return a.id, a.text, a.docid limit 5;
```

Explain logical/physical show SQL.

DuckDB's `rowid` column is mapped to `_ID`


adsharma:
TODO: add tests

PR #147, opened 2026-01-10
make datasets CLI importable

aheev:
Closes https://github.com/LadybugDB/ladybug/issues/86

PR #153, opened 2026-01-17
tests: make datatests cli importable v2

aheev:
This is a continuation PR for https://github.com/LadybugDB/ladybug/pull/147

Changes:
- fixed rem datasets missed by antigravity
- fixed storage path handling in `graph-std` datasets
- fixed bindings' test failures caused by prev PR. CI couldn't catch them as they are not run

### DIFF
--- a/tools/python_api/test/conftest.py
+++ b/tools/python_api/test/conftest.py
@@ -92,9 +92,11 @@ def init_long_str(conn: lb.Connection) -> None:
     conn.execute("CREATE REL TABLE knowsLongString (FROM personLongString TO personLongString, MANY_MANY)")
     conn.execute(f'COPY knowsLongString FROM "{LBUG_ROOT}/dataset/long-string-pk-tests/eKnows.csv"')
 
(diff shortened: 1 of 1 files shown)


aheev:
@adsharma PTAL

aheev:
> Looks good. Minor improvements in a future cleanup:
> 
> ```
> The regex-based file matching in common.js and conftest.py uses:
> const dataFileRegex = new RegExp(`"([^"]+\\.(${dataFileExtension.join('|')}))"`, "gi");
> This pattern matches any file with allowed extensions without path traversal protections (e.g., ../../../etc/passwd.csv). However, since these are test files running in controlled environments, the risk is limited.
> 
> Recommendations:
> 1. Add path normalization and validation before file operations
> 2. Use path.resolve() and verify paths stay within expected directories
> 3. Consider adding allowlist validation for dataset paths
> ```

tracking [here](https://github.com/LadybugDB/ladybug/issues/155)

Issue #168, opened 2026-02-05
Optimization:  Performing  DETACH-DELETE on a large-scale graph will consume a very long time


Mrhs121:
### Description

I created a graph locally with 10 million vertices and 50 million edges. When I was executing MATCH (n:User) where n.id <9999999 DETACH DELETE n, I waited for at least an hour but it still didn't finish. There are 9,999,999 vertices that meet the condition n.id <9999999, and the number of edges associated with them is 9,999,999 *5.

adsharma:
@Mrhs121 that'll be great. If you want to pursue the PR further, this could help:

```
--- a/src/processor/processor_task.cpp
+++ b/src/processor/processor_task.cpp
@@ -29,6 +29,10 @@ void ProcessorTask::run() {
     auto resultSet =
         sink->getResultSet(storage::MemoryManager::Get(*executionContext->clientContext));
     taskRoot->ptrCast<Sink>()->execute(resultSet.get(), executionContext);
+    // It's not super clear why this conditional is needed. Without it, issue.2589 fails
+    if (taskRoot->getOperatorType() == PhysicalOperatorType::RESULT_COLLECTOR) {
+        taskRoot->finalize(executionContext);
+    }
 }
```

The finalize() call is necessary to flush the last batch which could be smaller than `BATCH_SIZE`. However, adding the call breaks other tests as noted in the comment.

Perhaps its possible to use some other mechanism to handle the last batch that doesn't break tests.

A second optimization I was thinking was morsel driven parallelism. Multiple threads execute the delete in parallel. But it's complex because of potential dependencies. We'll need a dependency analysis pass that proves that it's parallelizable and then do it.

PR #169, opened 2026-02-05
Implement detachDeleteBatch()

adsharma:
Related to: #168


adsharma:
with batch_size=1024, I get ~70 secs for deleting 10 million nodes.

Also there is a bug somewhere that causes it to delete ignoring the filter condition. The test failures are legit.

adsharma:
Deprioritizing this work. The PR needs to be rewritten.

PR #223, opened 2026-02-23
implement semi-mask filtering in arrow table scan

aheev:
Right now, each arrow table batch scan returns all the rows ignoring the scanState.semiMask which is very inefficient. Implemented filtering in each scan by using this mask. Also made minor optimizations in condition checks

Tests:
To be added


adsharma:
Merged via #238 

Please send me a new PR to use the `applySemiMaskFilter()` from columnar base if there is no reason to carry a nearly identical copy in the arrow code.

And then we can work on verifying performance with and without semi masks.

PR #243, opened 2026-02-26
Refactor UUID

aheev:
This PR merges structs uuid and UUID. There's no point in having a separate class just to store static helper methods


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/22450939566

PR #264, opened 2026-03-03
optimize arrow batch scan

aheev:
- improve applySemiMaskFilter perf by using lower_bound instead of linear search
- populate outputToArrowColumnIdx only once table scan

### DIFF
--- a/src/storage/table/node_table.cpp
+++ b/src/storage/table/node_table.cpp
@@ -878,10 +878,10 @@ void NodeTable::applySemiMaskFilter(const TableScanState& state, row_idx_t start
             auto selectedPos = selVector.getSelectedPositions();
             auto itr = selectedPos.begin();
             auto end = selectedPos.end();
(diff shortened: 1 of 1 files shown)


src/storage/table/node_table.cpp
```diff
@@ -865,20 +867,35 @@ void NodeTable::applySemiMaskFilter(const TableScanState& state, row_idx_t start
...
-            auto temp = arr[j] - startOffset;
-            if (selVector[i] < temp) {
-                ++i;
-            } else if (selVector[i] > temp) {
-                ++j;
-            } else {
+        sel_t numSelectedValues = 0;
+
+        if (selVector.isUnfiltered()) {
+            for (size_t j = 0; j < arr.size(); ++j) {
```
adsharma: nit: since you removed `i`, perhaps rename `j` -> `i`?

Issue #180, opened 2026-02-14
Bug: Running Data Manipulation Clauses yields different results depending on how often database is opened and closed


andreasdamm-shure:
### Ladybug version

0.14.3

### What operating system are you using?

Windows 11 and Amazon Linux 2023

### What happened?

Running the given queries should delete all nodes and rels in the database. That behavior is observed when only opening the database once after creation, running the queries and then closing it.

When opening the database for each node to be modified and then closing it, many nodes are left undeleted but all the rels are gone.

### Are there known steps to reproduce?

Unzip [undead-nodes-bug.zip](https://github.com/user-attachments/files/25309485/undead-nodes-bug.zip) and run npm install.

Running "npm run good" creates undead-nodes.kz which when observed with the CLI contains no nodes.

Running "npm run bad"  creates undead-nodes.kz which when observed with the CLI contains many nodes but no rels.

andreasdamm-shure:
Is there a fix for this issue, yet?

PR #289, opened 2026-03-13
fix: Two MVCC bugs in the CSR storage layer

adsharma:
Fixes: #180 

The test is too expensive for CI. Will disable after an initial run.

### DIFF
--- a/test/storage/detach_delete_test.cpp
+++ b/test/storage/detach_delete_test.cpp
@@ -1,3 +1,5 @@
+#include <cstdlib>
+
 #include "common/types/value/value.h"
 #include "graph_test/private_graph_test.h"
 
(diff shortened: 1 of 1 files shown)


test/storage/detach_delete_test.cpp
```diff
@@ -0,0 +1,116 @@
+#include "common/types/value/value.h"
+#include "graph_test/private_graph_test.h"
+
+namespace lbug {
+namespace testing {
+
+// Reproduces the "bad" case from pristine.js: open a fresh Database+Connection on every
```
aheev: I believe you are plannning to remove this comment

aheev:
Another issue I have noticed is if I don't `createDBAndConn` every itr of the for loop, the result is

```
2533: countEdgesT: COUNT_STAR()
2533: 176
2533: 
2533: countEdgesB: COUNT_STAR()
2533: 176
```

PS: also `conn.reset();  // database.reset();` is not required as `createDBAndConn` already resets conn and database

PR #332, opened 2026-03-26
feat: non-blocking concurrent checkpoint (WAL rotation + MVCC snapshot)

loganpowell:
## Summary

This PR ports Vela-Engineering/kuzu's concurrent multi-writer checkpoint implementation
into ladybug. The implementation was developed by Aikins Laryea (Vela Partners) across
4 phases plus a read-only lock fix; the full diff is 2,360 lines across 45 files.

**Before**: checkpoint drains all transactions (reads + writes) and blocks new writes
until the entire checkpoint completes.

**After**: checkpoint only drains write transactions, releases the write gate as soon
as the WAL is rotated, and reads continue concurrently via MVCC snapshot isolation.
Shadow pages are applied with per-page CAS locking so optimistic readers detect updates.

### Core mechanism
- **WAL rotation**: at checkpoint start, the active `.wal` is atomically renamed to
  `.wal.checkpoint` (frozen WAL). New write transactions immediately create a fresh
  active WAL, isolated from the checkpoint.
- **Write gate release**: once WAL is rotated the gate is released; storage
  materialization runs concurrently with new writers via per-node-group locking.
- **MVCC catalog snapshots**: `Catalog::serializeSnapshot(snapshotTS)` and
  `CatalogSet::serializeSnapshot(snapshotTxn)` use a pinned snapshot timestamp for
  consistent serialization after the gate is released.
- **Epoch-based change tracking**: `Table::hasChanges` (bool) replaced by
  `changeEpoch` (atomic uint64). `captureChangeEpochs()` snapshots per-table
  watermarks under the write gate; `Table::checkpoint()` uses watermarks to skip
  tables that haven't changed since the last gate.

### Thread safety
- `StorageManager::mtx` upgraded from exclusive `std::mutex` to
  `mutable std::shared_mutex` (reads take shared lock, DDL takes unique lock).
- `NodeTable::schemaMtx` protects the `columns` vector during concurrent
  checkpoint vacuum.
- `PageManager::version` and `Catalog::version` use `std::atomic<uint64_t>`
  with a stable `lastCheckpointVersion` baseline.
- `ShadowFile::applyShadowPages` now calls `BufferManager::updateFrameIfPageIsInFrame`
  (CAS-loop locked variant) instead of the lock-free variant.

### WAL recovery
- `WALReplayer::replay` now handles both frozen (`.wal.checkpoint`) and active
  (`.wal`) WAL files independently via `replayFrozenWAL` / `replayActiveWAL`.

### Read-only databases
- File lock is skipped when `readOnly=true`, enabling concurrent read-only opens.

### Apple clang compatibility
- Explicit constructors added to `SegmentCheckpointState` and `LazySegmentData`.

## Testing
- Updated `test/transaction/checkpoint_test.cpp` to use the new
  `logCheckpointAndApplyShadowPages(bool walRotated)` signature, WAL rotation-aware
  flaky checkpointer classes, and `ShadowFileDatabaseIDMismatchExistingDB` now
  validates the frozen WAL (`.wal.checkpoint`) rather than the active WAL.
- CI workflow at `embed-graph-db/.github/workflows/ladybug-concurrent-writes-ci.yml`
  covers build (ubuntu + macos), full ctest suite, checkpoint/concurrent-write test
  filter, and benchmark comparison (blocks on >5% single-writer regression or
  zero read-during-checkpoint throughput).

## Attribution
Concurrent checkpoint implementation: Aikins Laryea, Vela Partners
([Vela-Engineering/kuzu](https://github.com/Vela-Engineering/kuzu), commits 3ad9071..d298e9c)

Ported and adapted for ladybug (`lbug` namespace, `mainStorageManager` field,
2-arg `applyShadowPages` API, `std::format`, graph catalog support) by Logan Powell.

### DIFF
--- a/.github/workflows/concurrent-writes-ci.yml
+++ b/.github/workflows/concurrent-writes-ci.yml
@@ -24,8 +24,8 @@ env:
 
 jobs:
   # ─── Format check ─────────────────────────────────────────────────────────
-  # Mirrors the upstream Docker script (run-clang-format-docker.sh) using the
-  # same clang-18 version available on Ubuntu without requiring Docker-in-Docker.
+  # Uses the upstream run-clang-format-docker.sh script (Alpine + clang18-extra-tools)
+  # so CI uses the exact same binary as local dev — avoids Ubuntu apt clang-18 patches.
   format:
     name: clang-format-18 check
     runs-on: ubuntu-latest
@@ -34,11 +34,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0   # need history to diff against main
-
-      - name: Install clang-format-18
-        run: |
-          sudo apt-get update -qq
-          sudo apt-get install -y clang-format-18
 
       - name: Check formatting of changed C++ files
         run: |
@@ -49,9 +44,7 @@ jobs:
             exit 0
           fi
           echo "Checking $(echo "$CHANGED" | wc -l | tr -d ' ') files..."
-          python3 scripts/run-clang-format.py \
-            --clang-format-executable clang-format-18 \
-            $CHANGED
+          bash scripts/run-clang-format-docker.sh $CHANGED
 
   # ─── Build (branch) ───────────────────────────────────────────────────────
   build:
--- a/.github/workflows/concurrent-writes-ci.yml
+++ b/.github/workflows/concurrent-writes-ci.yml
@@ -7,13 +7,9 @@ name: ladybug concurrent-writes CI
 #   - Benchmark comparison → PR comment + GitHub Step Summary
 
 on:
-  push:
-    branches: [feat/concurrent-writes]
-    paths:
-      ["src/**", "benchmarks/**", ".github/workflows/concurrent-writes-ci.yml"]
-  pull_request:
-    paths:
-      ["src/**", "benchmarks/**", ".github/workflows/concurrent-writes-ci.yml"]
+  schedule:
+    - cron: '0 2 * * *'  # 02:00 UTC nightly on feat/concurrent-writes
+  workflow_dispatch:       # allow manual trigger any time
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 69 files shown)


adsharma:
Also I misread the duckdb docs around "single writer" model. Here's the relevant paragraph:

> DuckDB supports concurrency within a single process according to the following rules. As long as there are no write conflicts, multiple concurrent writes will succeed. Appends will never conflict, even on the same table. Multiple threads can also simultaneously update separate tables or separate subsets of the same table. Optimistic concurrency control comes into play when two threads attempt to edit (update or delete) the same row at the same time. In that situation, the second thread to attempt the edit will fail with a conflict error.

Good news: this PR doesn't break the duckdb concurrency model. I have no objections to it at a conceptual level.

I have some disagreements with the duckdb folks about VFS level concurrency. I think they're stricter than they need to be to the point where a WAL replication setup is impossible on all platforms. But that's a separate topic.

For this particular PR, we just need to work through the 3 review comments and a post-merge MVCC testing plan to flush out any bugs.

loganpowell:
Beautiful. Thank you sir, for your great work @adsharma and for the careful guidance.

GlitterKill:
> Overall: The "do not block readers while checkpointing" part is more interesting than concurrent write throughput. I don't think the ladybugdb user base cares about concurrent write throughput.
> 
> Historically, this project is very closely related to DuckDB, so I'd be careful about deviating substantially from DuckDB's concurrency model. DuckDB supports single writer, but allows multiple-readers during checkpointing.
> 
> https://duckdb.org/docs/stable/connect/concurrency
> 
> * The CI doesn't need clang-format (already handled in ci-workflow.yml) and this doesn't need to run on every PR.
> * Running it as a part of nightly scheduled CI is a good idea
> 
> We should also do some antithesis based runs to flush out bugs arising out of this PR or pre-existing issues. Let me know if it's something you're excited about taking on after this is merged.

I would be interested in concurrent write throughput.  :)

adsharma:
Let's use #308 for the discussion.

PR #358, opened 2026-03-31
ci: use ubuntu 22.04 for wider compatibility

adsharma:


.github/workflows/java-workflow.yml
```diff
@@ -20,14 +20,20 @@ env:
...
     env:
       LBUG_LINUX_VARIANT: compat
       LBUG_JAVA_PRECOMPILED_DIR: ${{ github.workspace }}/precompiled-liblbug
       LBUG_JAVA_PRECOMPILED_LIB_PATH: ${{ github.workspace }}/precompiled-liblbug/liblbug.a
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install GCC 13
+        run: |
+          sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
```
antonymarion: @adsharma it means that people on ubuntu 22.04 for instance not adding this ppa to benefit the last gcc g++ libs will have the following error "Error: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.32' not found" since they will be up to date "libstdc++6 is already the newest version (12.3.0-1ubuntu1~22.04.3)." but not regarding the ppa you used right ?

Not a big deal for you?

PR #381, opened 2026-04-14
fix checkpoint segment fault

### LINKED ISSUE #380: Bug: NodeTable::checkpoint() segment fault on retry due to non-exception-safe destructive column move
### Ladybug version

_No response_

_No response_

### What happened?

Image

When a database with pending WAL records (e.g., `ALTER TABLE ADD COLUMN`) is opened and `checkpoint` is executed, a segfault (SIGSEGV) occurs during `Database::~Database` when
  `forceCheckpointOnClose` triggers a second checkpoint.

  ### Crash Stack

  EXC_BAD_ACCESS (code=1, address=0x20)

  frame #0: LogicalType::LogicalType(other=0x20)           // copying from invalid pointer
  frame #3: NodeGroup::scanAllInsertedAndVersions()
  frame #4: NodeGroup::checkpointInMemAndOnDisk()
  frame #5: NodeGroup::checkpoint()
  frame #6: NodeGroupCollection::checkpoint()
  frame #7: NodeTable::checkpoint()
    ...
  frame #13: Database::~Database()

### RC

NodeTable::checkpoint() performs a destructive std::move on the columns vector (vacuuming dropped columns) before calling nodeGroups->checkpoint(). If nodeGroups->checkpoint() throws, the
  function exits with:

  - columns already moved (e.g., shrunk from 18 → 17 elements)
  - tableEntry column IDs not vacuumed (still referencing original IDs up to 17)
  - lastCheckpointedEpoch not updated

  When Database::~Database calls forceCheckpointOnClose, NodeTable::checkpoint() re-enters, and the non-vacuumed column IDs index out of bounds into the already-shrunk columns vector,
  producing a garbage pointer that dereferences to a segfault.

### Fix

  Defer the destructive column move until after nodeGroups->checkpoint() succeeds. Build raw Column* pointers from the existing columns vector without moving ownership. Only perform the
  std::move and vacuumColumnIDs once checkpoint is confirmed successful. This makes the function exception-safe: on failure, columns and catalog state remain consistent for any subsequent
  retry.

Open a database that has uncommitted WAL records containing ALTER TABLE ADD COLUMN

./lbug test.lbug
and execute checkpoint

Mrhs121:
close https://github.com/LadybugDB/ladybug/issues/380

### DIFF
--- a/src/storage/table/node_table.cpp
+++ b/src/storage/table/node_table.cpp
@@ -658,20 +658,19 @@ bool NodeTable::checkpoint(main::ClientContext* context, TableCatalogEntry* tabl
     if (effectiveEpoch <= lastCheckpointedEpoch) {
         return false;
     }
-    // Build column IDs and pointers under unique_lock to protect from concurrent readers.
+    // Build column IDs and raw pointers for checkpoint WITHOUT moving columns yet.
+    // The destructive column move is deferred until after nodeGroups->checkpoint() succeeds.
+    // This ensures exception safety: if checkpoint fails, columns remain in their original
+    // state and a subsequent checkpoint (e.g., from Database::~Database) won't crash due to
+    // an inconsistency between the columns vector and the catalog's column IDs.
     std::vector<column_id_t> columnIDs;
     std::vector<Column*> checkpointColumnPtrs;
     {
         std::unique_lock schemaLck{schemaMtx};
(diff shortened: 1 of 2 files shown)


adsharma:
The fix sounds right. Could you also add a test?

Mrhs121:
> The fix sounds right. Could you also add a test?
Done

adsharma:
Thank you!

Issue #378, opened 2026-04-13
Bug: Duplicate Rel IDs


andreasdamm-shure:
### Ladybug version

0.15.1

### What operating system are you using?

MacOS Tahoe 26.4

### What happened?

Run the following queries in the CLI creating a new database file:

```cypher
create node table D(hid string primary key);

create node table E(
    id string default gen_random_uuid(),
    t uint8,
    r boolean default false,
    p int64,
    primary key (id)
);

create node table R(id string primary key);

create rel table RB(from E to R);

create rel table T(from R to D);

create rel table PB(from E to D, many_one);

create rel table RL(from E to D, many_one);

create rel table PO(from E to E);

begin transaction;

unwind ["p1", "c1", "c2"] as id
create (:D { hid: id });

unwind [
    {
        t: 0,
        p: 0,
        rhid: null,
        phid: "p1",
        rs: [
            {id: "r1", hid: "p1"}
        ]
    },
    {
        t: 2,
        p: 1,
        rhid: null,
        phid: "p1",
        rs: [
            {id: "r2", hid: "c1"},
            {id: "r3", hid: "c2"}
        ]
    }
] as edetail
match (px:D { hid: edetail.phid })
optional match (rl:D { hid: edetail.rhid })
create 
    (x:E { t: edetail.t, p: edetail.p })-[:PB]->(px),
    (x)-[:RL]->(rl)
with x, edetail
unwind cast(edetail.rs as struct(id string, hid string)[]) as rdetail
match (dv:D { hid: rdetail.hid })
create (x)-[:RB]->(:R { id: rdetail.id })-[:T]->(dv);

match (x:E)
where x.p is not null
with list_sort(collect(x.p)) as ps
unwind ps as p
match (x:E { p: p })-[:PB]->(pf:D)
match (before:E)-[:PB]->(pf)
where 
    (before.p is null or before.p < p) and
    case x.t
        when 0 then before.t = 0
        when 1 then before.t = 0 or before.t = 1
        else true
    end
create (before)-[:PO]->(x);

match (x:E)
where x.p is not null
set x.p = null;



match (x:E { r: false })-[:RB]->(rq:R)-[:T]->(target:D)
where not exists { match ()-[]->(x) }
with x, collect({ hid: target.hid, rid: rq.id }) as ts
match (x)-[:PB]->(pf:D)
optional match (x)-[:RL]->(rl:D)
set x.r = true
return x, ts, pf, rl;

commit;
```

[...]

PR #389, opened 2026-04-17
Initialize ScanMultiRelTable state before first scan

adsharma:
Related: #378 

Initialize the first rel-table scan state when a multi-rel scanner starts scanning a new bound node. Without this, the operator could reuse stale scan state and direction from a previous table, causing incorrect relationship direction behavior in subquery edge-existence checks.


adsharma:
Don't think this is necessary. Covered by #398

PR #397, opened 2026-04-18
add codeql workflow

aheev:
enable codeQL scanning on core and actions. runs every friday midnight


aheev:
@adsharma could you PTAL?

PR #402, opened 2026-04-21
optimize codeQL run

aheev:
improve codeQL runtime by creating a minimal build for codeQL analysis


aheev:
@adsharma could you PTAL?

PR #408, opened 2026-04-21
ci: musl use the same uid/gid as ubuntu

adsharma:


magyargergo:
Seems `cmake` couldn't find openssl

adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/24739701974/job/72375896557

Issue #393, opened 2026-04-17
Python binding improvements


adsharma:
Some architectural directions

PR #418, opened 2026-04-24
python: use the dual backend structure

adsharma:
Related to: #393 

* Update the submodule
* Fixup scripts that create sdist


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/24902274618

PR #425, opened 2026-04-27
Fix eviction queue race in buffer manager

### LINKED ISSUE #350: ci: LargeAggregateLeakTest is flaky
https://github.com/LadybugDB/ladybug/actions/runs/23691085207/job/69018677897

shows

 Test #1805: e2e_test_agg~hash_leak.LargeAggregateLeakTest

started, but never passed.

When this happens, the CI hangs for 5 hours.

adsharma:
Fixes: #350 

Earlier fix was insufficient. There is another race.


adsharma:
This commit fixes a race condition where two threads could simultaneously handle the same EVICTED page, causing inconsistent state between the eviction queue and page states. The fix introduces atomic operations and proper versioning to ensure thread safety.

## **Race Condition Analysis**

The issue occurs when multiple threads scan the eviction queue and encounter the same evicted page candidate. Previously, the clear() operation wasn't atomic with respect to the expected candidate value, allowing race conditions where one thread could clear a candidate while another was still processing it.

## **Key Changes**

### 1\. Atomic tryClear Method

Added EvictionQueue::tryClear() which uses compare\_exchange\_strong to only clear the candidate if it matches the expected value buffer\_manager.cpp:64-76 . This prevents clearing a candidate that another thread has already modified.

### 2\. Version Increment in resetToEvicted()

Modified PageState::resetToEvicted() to increment the version when transitioning to EVICTED state page\_state.h:87-94 . This ensures that stale reads can be detected after a page has been evicted and potentially reloaded.

### 3\. Safer Eviction Logic

In evictPages(), when handling evicted candidates, the code now:

* Uses tryClear() instead of clear() to ensure atomicity  
* Calls resetToEvicted() on the page state to properly increment the version buffer\_manager.cpp:288-298

In tryEvictPage(), reordered operations to clear the eviction queue before resetting the page state buffer\_manager.cpp:445-455 .

## **Thread Safety Improvements**

The fix ensures that:

1. Only one thread can successfully clear a specific eviction candidate  
2. Page state versions are properly incremented during eviction  
3. The eviction queue and page states remain consistent even under concurrent access

This prevents scenarios where two threads could think they both successfully evicted the same page, leading to memory leaks or corruption.

## **Notes**

The version increment in resetToEvicted() is particularly important because it prevents optimistic reads from returning stale data after a page has been evicted and potentially reloaded by another thread buffer\_manager.h:168-172 .

PR #450, opened 2026-05-03
fix(file_system): improve lock error message by adding PID and OS error details

ksmeltzer:
## Description

Improves the lock error message when a process cannot acquire a lock on the database files by adding the blocking process ID (PID) and OS error details to the POSIX lock failure messages using `fcntl(F_GETLK)`. Adds `posixErrMessage()` or `GetLastError()` details directly to the lock failure messages to make debugging easier.

This fixes confusing lock errors where users could not identify what process was holding the LadybugDB database lock.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation Update

## Checklist

- [ ] I have changed storage version if on disk format has changed.
- [ ] I have updated the documentation (if needed).

### DIFF
--- a/src/common/file_system/local_file_system.cpp
+++ b/src/common/file_system/local_file_system.cpp
@@ -145,7 +145,8 @@ std::unique_ptr<FileInfo> LocalFileSystem::openFile(const std::string& path, Fil
         fl.l_len = 0;
         int rc = fcntl(fd, F_SETLK, &fl);
         if (rc == -1) {
-            if (errno == EAGAIN || errno == EACCES) {
+            int original_errno = errno;
+            if (original_errno == EAGAIN || original_errno == EACCES) {
                 struct flock get_fl {};
                 memset(&get_fl, 0, sizeof get_fl);
                 get_fl.l_type = flags.lockType == FileLockType::READ_LOCK ? F_RDLCK : F_WRLCK;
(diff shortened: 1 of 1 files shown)


ksmeltzer:
All checks have passed! Could a maintainer please take a look when you have a moment? cc @benjaminwinger

src/common/file_system/local_file_system.cpp
```diff
@@ -143,9 +145,27 @@ std::unique_ptr<FileInfo> LocalFileSystem::openFile(const std::string& path, Fil
...
+                    if (get_fl.l_type != F_UNLCK) {
+                        throw IOException(
+                            "Could not set lock on file : " + fullPath + " (Lock is held by PID " +
+                            std::to_string(get_fl.l_pid) + ")\n" +
+                            "See the docs: https://docs.ladybugdb.com/concurrency for more "
+                            "information.");
+                    }
+                }
+            }
+            throw IOException("Could not set lock on file : " + fullPath +
```
adsharma, line 167: If `F_GETLK` fails, this will report the errno from that call instead of the original errno from `F_SETLK`.

adsharma (changes requested):
LGTM - just one comment.

src/common/file_system/local_file_system.cpp
ksmeltzer: Good catch! I've updated the logic to stash `errno` immediately after `F_SETLK` fails, and then restore it back before generating the generic `posixErrMessage()` at the bottom. This ensures we don't accidentally swallow the original reason if `F_GETLK` fails or clears the error state. Thanks for reviewing!

adsharma:
Thank you for your contribution!

PR #453, opened 2026-05-04
Fix inline property filter support for ANY graph queries

### LINKED ISSUE #451: Bug: open type graph error
### Ladybug version

v0.15.3

opensuse tumbleweed linux

### What happened?

Hello,

I've just downloaded latest lbug explorer, via ghcr docker image, to try the open type graphs feature (https://github.com/LadybugDB/ladybug/pull/152) 

tried running the example query:
```cypher
CREATE GRAPH mygraph ANY; 
USE GRAPH mygraph; 
CREATE (:User {name: 'Alice', age: 30})-[:LivesIn]->(:City {name: 'New York', population: 8000000}); 
match (a) return a.*
```

and got: 'Unsupported type: JSON'

_No response_

aheev:
Two bugs prevented MATCH (a:Label { prop: val }) from working in ANY graphs:

1. JSON string serialization mismatch in ColumnCaster (scan_table.cpp): yyjson_val_write() serialized JSON strings as '"Alice"' (with JSON quote delimiters), but CAST(STRING → JSON) on the RHS produced 'Alice' (no delimiters). Numbers accidentally worked because both sides produced '30'. Fix: add jsonExtractScalarToString() which uses yyjson_get_str() for string JSON values (returns raw string content), falling back to jsonToString() for numbers and booleans.

2. Cardinality estimator crash (cardinality_estimator.cpp): getTableStatsIfPossible() called entry->getColumnID(propertyName) without checking whether the property actually exists as a column. In ANY graphs, properties like 'age' are stored as JSON in the 'data' column — not as individual columns — so nameToPropertyIDMap.at('age') threw unordered_map::at. Fix: add entry->containsProperty(propertyName) guard before getColumnID call.

Also updates the QueryDynamicProperties test expectation: with the scan fix, string properties now return unquoted values (e.g., Alice instead of "Alice"), consistent with regular graph property access.

Adds four new test cases to test/test_files/graph/any.test:
- InlineIntPropertyFilter
- InlineStringPropertyFilter
- InlinePropertyFilterReturnStar
- InlinePropertyFilterNoMatch

Closes https://github.com/LadybugDB/ladybug/issues/451

### DIFF
--- a/src/common/json_utils.cpp
+++ b/src/common/json_utils.cpp
@@ -702,7 +702,7 @@ std::string jsonExtractToString(const JsonWrapper& wrapper, uint64_t pos) {
     return jsonToString(yyjson_arr_get(yyjson_doc_get_root(wrapper.ptr), pos));
 }
 
-std::string jsonExtractToString(const JsonWrapper& wrapper, std::string path) {
+static yyjson_val* jsonNavigatePath(const JsonWrapper& wrapper, const std::string& path) {
     std::vector<std::string> actualPath;
     for (auto i = 0u, prvDelim = 0u; i <= path.size(); i++) {
         if (i == path.size() || path[i] == '/' || path[i] == '.') {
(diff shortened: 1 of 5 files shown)


adsharma:
This handles only the node filters. I was working on a more generic fix that handles both node and edge filters:

```
MATCH (a:User)-[e:livesIn {since: 2020}]->(c:City) RETURN e.since;
```

Please review the incoming PR. Main delta is it touches map_extend.cpp

aheev:
fixed by https://github.com/LadybugDB/ladybug/pull/454

PR #454, opened 2026-05-04
Fix open graph dynamic property filters

### LINKED ISSUE #451: Bug: open type graph error

adsharma:
Fixes: #451 

Open type graph dynamic properties are stored in the system data `JSON` column, but filtered scans were treating them like physical table columns. This made pattern predicates and WHERE predicates on labels such as User.age call property metadata lookups that do not exist for _nodes/_edges, which surfaced as unordered_map::at failures instead of behaving like a declared node or edge table.

Route missing dynamic properties through JSON extraction during scan planning, apply the same casting path for relationship scans, and avoid cardinality estimation lookups for properties that are not physical columns. Add ANY graph regressions for node and relationship property filters so open graphs match the behavior users get after creating explicit node and edge tables.

### DIFF
--- a/src/common/json_utils.cpp
+++ b/src/common/json_utils.cpp
@@ -702,7 +702,9 @@ std::string jsonExtractToString(const JsonWrapper& wrapper, uint64_t pos) {
     return jsonToString(yyjson_arr_get(yyjson_doc_get_root(wrapper.ptr), pos));
 }
 
-std::string jsonExtractToString(const JsonWrapper& wrapper, std::string path) {
+namespace {
+
+yyjson_val* extractJsonByPath(const JsonWrapper& wrapper, const std::string& path) {
     std::vector<std::string> actualPath;
     for (auto i = 0u, prvDelim = 0u; i <= path.size(); i++) {
         if (i == path.size() || path[i] == '/' || path[i] == '.') {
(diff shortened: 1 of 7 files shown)


test/test_files/graph/any.test
```diff
@@ -33,6 +33,26 @@
 ---- 1
 "Alice"|"New York"
 
+-LOG QueryDynamicNodePropertyInPatternFilter
+-STATEMENT MATCH (u:User {age: 30}) RETURN u.name, u.age
+---- 1
+"Alice"|30
```
aheev: Shouldn’t we return a plain string instead of a quoted one, just like the normal graphs?

aheev: Also a test involving string prop filters would help

aheev (changes requested):

test/test_files/graph/any.test
adsharma: Both good suggestions. Yes - the ideal of end users can't detect any differences between strictly typed vs open type graphs at the query level is worth going for.

src/common/json_utils.cpp
```diff
@@ -731,6 +731,38 @@ std::string jsonExtractToString(const JsonWrapper& wrapper, std::string path) {
     return jsonToString(ptr);
 }
 
+std::string jsonExtractScalarToString(const JsonWrapper& wrapper, std::string path) {
+    std::vector<std::string> actualPath;
```
aheev: nit: can this code be de-dupped?

aheev (approved):
LGTM except one nit

Issue #183, opened 2026-02-14
Optimize Arrow Memory Tables


adsharma:
## Background
[Commit `23d0e8aa0`](https://github.com/LadybugDB/ladybug/commit/23d0e8aa0) introduced Arrow memory-backed tables (`ArrowNodeTable`) to enable zero-copy data sharing with Arrow ecosystem. However, the current implementation has performance limitations compared to native Ladybug tables.

## Current Implementation Issues

The current `ArrowNodeTable` implementation (see `src/storage/table/arrow_node_table.cpp`):
1. **Converts all Arrow data upfront** into `std::vector<std::unique_ptr<common::Value>>` (see `readArrowTableData()`)
2. **Returns one row at a time** during scan (row-at-a-time processing)
3. **Does NOT support semi-masks** for filtering during scan
4. **Does NOT utilize selection vectors** for efficient batch filtering
5. **Missing SIP (Sideways Information Passing)** optimization for joins

### 1. Support SIP (Sideways Information Passing) / Semi-Mask Filtering

**Problem:** Arrow tables currently ignore semi-masks set on `TableScanState::semiMask`, which means:
- Hash joins cannot push down semi-join filters to Arrow table scans
- Subquery unnesting cannot benefit from semi-mask-based filtering
- Missing significant join optimization opportunities

**Implementation:**
- In `ArrowNodeTable::scanInternal()`, check if `scanState.semiMask` is enabled
- Apply semi-mask filtering similar to `NodeGroup::applySemiMaskFilter()` (see `src/storage/table/node_group.cpp:186-213`)
- Filter rows based on `semiMask->range(startOffset, endOffset)` before returning data
- Skip rows that don't match the semi-mask to avoid unnecessary data conversion

**Key files:**
- `src/storage/table/arrow_node_table.cpp`
- `src/include/common/mask.h` (SemiMask interface)
- `src/storage/table/node_group.cpp` (reference implementation)

### 2. Support Selection Vectors

**Problem:** The current implementation always returns a selection vector with size 1 (`setSelSize(1)`), ignoring the vectorized processing capabilities of Ladybug.

**Implementation:**
- Process data in batches (up to `DEFAULT_VECTOR_CAPACITY` rows) instead of one row at a time
- Populate selection vector with valid row indices after applying filters
- Support predicate pushdown through selection vectors
- Handle both flat and unflat output vectors properly

**Key files:**
- `src/include/common/data_chunk/sel_vector.h` (SelectionVector interface)
[...]

PR #460, opened 2026-05-06
Support Arrow relationship table scans

adsharma:
Related: #183

### DIFF
--- a/src/include/storage/table/rel_table.h
+++ b/src/include/storage/table/rel_table.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <unordered_map>
+
 #include "catalog/catalog_entry/rel_group_catalog_entry.h"
 #include "storage/table/rel_table_data.h"
 #include "storage/table/table.h"
@@ -26,6 +28,15 @@ struct RelTableScanState : TableScanState {
     common::SelectionVector cachedBoundNodeSelVector;
 
     std::unique_ptr<LocalRelTableScanState> localTableScanState;
+
+    // Optional state used by Arrow-backed relationship tables. Keep it on the common scan state so
+    // a single multi-rel scan state can scan native, Parquet-backed, and Arrow-backed tables.
+    size_t arrowCurrentBatchIdx = 0;
+    size_t arrowCurrentBatchOffset = 0;
+    std::unordered_map<common::offset_t, common::sel_t> arrowBoundNodeOffsetToSelPos;
+    std::unique_ptr<common::ValueVector> arrowSrcKeyVector;
+    std::unique_ptr<common::ValueVector> arrowDstKeyVector;
+    bool arrowScanCompleted = true;
 
     RelTableScanState(MemoryManager& mm, common::ValueVector* nodeIDVector,
         std::vector<common::ValueVector*> outputVectors,
--- a/src/storage/table/arrow_node_table.cpp
+++ b/src/storage/table/arrow_node_table.cpp
@@ -50,32 +50,6 @@ ArrowNodeTable::~ArrowNodeTable() {
     }
 }
 
-void ArrowNodeTableScanState::setToTable(const transaction::Transaction* transaction, Table* table_,
-    std::vector<common::column_id_t> columnIDs_,
-    std::vector<ColumnPredicateSet> columnPredicateSets_, common::RelDataDirection) {
-    TableScanState::setToTable(transaction, table_, columnIDs_, std::move(columnPredicateSets_));
-
-    // populating outputToArrowColumnIdx for use in ArrowNodeTable scanInternal
-    this->outputToArrowColumnIdx.assign(columnIDs.size(), -1);
-    auto tableCatalogEntry = table->cast<ArrowNodeTable>().getCatalogEntry();
-
-    for (size_t col = 0; col < columnIDs.size(); ++col) {
-        const auto columnID = columnIDs[col];
-
-        if (columnID == common::INVALID_COLUMN_ID || columnID == common::ROW_IDX_COLUMN_ID) {
-            continue;
-        }
-
-        for (common::idx_t propIdx = 0; propIdx < tableCatalogEntry->getNumProperties();
-             ++propIdx) {
-            if (tableCatalogEntry->getColumnID(propIdx) == columnID) {
-                this->outputToArrowColumnIdx[col] = static_cast<int64_t>(propIdx);
-                break;
-            }
-        }
-    }
-}
-
 void ArrowNodeTable::initializeScanCoordination(const transaction::Transaction* transaction) {
     auto arrowScanSharedState =
         static_cast<ArrowNodeTableScanSharedState*>(tableScanSharedState.get());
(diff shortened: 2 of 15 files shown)


src/include/storage/table/columnar_node_table_base.h
```diff
@@ -17,6 +17,10 @@ struct ColumnarNodeTableScanState : NodeTableScanState {
     bool initialized = false; // Track if this scan state has been initialized for scanning
     size_t totalRows = 0;
     bool scanCompleted = false; // Track if this scan state has finished reading
+    size_t currentBatchIdx = static_cast<size_t>(common::INVALID_NODE_GROUP_IDX);
```
aheev: I don't see any usages of these members in parquet 🤔

src/storage/table/arrow_node_table.cpp
```diff
@@ -85,7 +59,7 @@ void ArrowNodeTable::initializeScanCoordination(const transaction::Transaction*
 
 void ArrowNodeTable::initScanState([[maybe_unused]] transaction::Transaction* transaction,
     TableScanState& scanState, [[maybe_unused]] bool resetCachedBoundNodeSelVec) const {
-    auto& arrowScanState = scanState.cast<ArrowNodeTableScanState>();
+    auto& arrowScanState = scanState.cast<ColumnarNodeTableScanState>();
```
aheev: My understanding is that Arrow and Parquet scans are fundamentally different. The difference can be attribute to their inherent storage. Because arrow tables sit in memory, we can completely own the parallelism. As for parquet, rowGroup size is not always the same. So, we can't distribute them uniformly across morsels everytime(when numMorsels != numRowGroups)

Take a look at getNextMorsel for example. arrow morsel size is constant(decided by us) where as for parquet, it is the rowGroup size. In each morsel, parquet_node_table runs another batching(size: 2048) internally

Maybe that's why separate scanStates were added to extend common columnarTableBase

PS: we can still make it consistent, but requires significant surgery at PhysicalOperator level

src/include/storage/table/columnar_node_table_base.h
```diff
@@ -17,6 +17,10 @@ struct ColumnarNodeTableScanState : NodeTableScanState {
...
+    size_t currentMorselStartOffset = 0;
+    size_t currentMorselEndOffset = 0;
+    std::vector<int64_t> outputToColumnarColumnIdx;
```
aheev: I think `outputToColumnarColumnIdx` belongs on the table class. Because it doesn't change across different morsels(scanStates) and can be populated in `Table.initScanCoordination`

src/include/storage/table/rel_table.h
```diff
@@ -27,6 +29,16 @@ struct RelTableScanState : TableScanState {
 
     std::unique_ptr<LocalRelTableScanState> localTableScanState;
 
+    // Optional state used by Arrow-backed relationship tables. Keep it on the common scan state so
+    // a single multi-rel scan state can scan native, Parquet-backed, and Arrow-backed tables.
+    size_t arrowCurrentBatchIdx = 0;
```
aheev, line 34: same comment as node table

src/storage/table/arrow_node_table.cpp
```diff
@@ -50,32 +50,6 @@ ArrowNodeTable::~ArrowNodeTable() {
     }
 }
 
-void ArrowNodeTableScanState::setToTable(const transaction::Transaction* transaction, Table* table_,
```
aheev, line 53: Thanks for removing this. This was not required in the first place(added by me 🙃 )

src/storage/table/arrow_node_table.cpp
```diff
@@ -98,6 +72,21 @@ void ArrowNodeTable::initScanState([[maybe_unused]] transaction::Transaction* tr
     }
 
     // Each scan state needs to be able to read data independently for parallel scanning
+    arrowScanState.outputToColumnarColumnIdx.assign(scanState.columnIDs.size(), -1);
+    auto tableCatalogEntry = getCatalogEntry();
+    for (size_t col = 0; col < scanState.columnIDs.size(); ++col) {
```
aheev: This should belong to initScanCoordination

test/api/arrow_rel_table_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+                    "CREATE REL TABLE arrow_rel_transfer(FROM arrow_rel_account TO "
+                    "arrow_rel_account);"
+                    "CREATE (:arrow_rel_account {id: 10})-[:arrow_rel_transfer]->"
+                    "(:arrow_rel_account {id: 20});");
+    ASSERT_TRUE(createNativeTables->isSuccess()) << createNativeTables->getErrorMessage();
+
+    auto result = conn->query("MATCH ()-[]->() RETURN count(*)");
+    ASSERT_TRUE(result->isSuccess()) << result->getErrorMessage();
+    ASSERT_EQ(result->getNext()->getValue(0)->getValue<int64_t>(), 4);
+}
```
aheev, line 152: Can we add one more to the mix: `native node tables, arrow rel tables` just to be complete ?

src/storage/table/arrow_node_table.cpp
adsharma: My inclination is to make mixed tables work (it's made easier by unifying the scan states), add tests and then specialize them for performance.

adsharma:
Changes:

  - Moved Arrow output-column mapping out of scan state into Arrow table helper methods.
  - Kept mixed native/Arrow/Parquet scan compatibility through shared cursor/key state where needed.
  - Fixed the Arrow rel scan cursor so it does not skip the first row for the next bound node.
  - Added ScanArrowRelTableOverNativeNodeTable coverage.

src/storage/table/arrow_node_table.cpp
adsharma: `initScanCoordination` does not have the operator-specific projected columnIDs; it only coordinates table-level morsel assignment. The Arrow column mapping depends on the current scan’s requested output columns, so I kept it derived at scan/init-scan time from table metadata rather than storing it as global coordination state.

adsharma:
- Removed Arrow-specific currentBatchIdx/currentMorselStartOffset/currentMorselEndOffset from ColumnarNodeTableScanState.
  - Moved those fields back onto ArrowNodeTableScanState.
  - Changed ScanNodeTable to create the correct scan-state type per current node table instead of forcing one columnar state to cover Arrow and Parquet. That preserves mixed-table behavior without leaking Arrow cursor state into the base.

adsharma:
@aheev - there are two unresolved comments that may require further discussion. Let's discuss them on an issue/discussion if necessary.

I want to close on the python + arrow bindings and share some sample code/notebooks.

aheev:
> @aheev - there are two unresolved comments that may require further discussion. Let's discuss them on an issue/discussion if necessary.
> 
> I want to close on the python + arrow bindings and share some sample code/notebooks.

sure we'll get back to them next week

PR #476, opened 2026-05-10
refactor parquet -> icebug-disk

### LINKED ISSUE #469: Feature: Add Icebug-disk spec impl

aheev:
Apart from renames
- added checks for mixed type tables
- added version checks
- updated  `storage = ` format
- fixed StorageManager defaulting to icebug-disk
- added doc

Closes https://github.com/LadybugDB/ladybug/issues/469

### DIFF
--- a/src/include/common/constants.h
+++ b/src/include/common/constants.h
@@ -86,6 +86,12 @@ struct StorageConstants {
 struct TableOptionConstants {
     static constexpr char REL_STORAGE_DIRECTION_OPTION[] = "STORAGE_DIRECTION";
     static constexpr char REL_STORAGE_OPTION[] = "STORAGE";
+
+    static constexpr std::string_view ICEBUG_DISK_PREFIX = "icebug-disk";
+
+    static bool isIceBugDiskStorage(const std::string& storage) {
+        return storage.starts_with(ICEBUG_DISK_PREFIX);
+    }
 };
 
 // Hash Index Configurations
--- a/src/include/common/constants.h
+++ b/src/include/common/constants.h
@@ -86,11 +86,12 @@ struct StorageConstants {
 struct TableOptionConstants {
     static constexpr char REL_STORAGE_DIRECTION_OPTION[] = "STORAGE_DIRECTION";
     static constexpr char REL_STORAGE_OPTION[] = "STORAGE";
+    static constexpr char STORAGE_FORMAT_OPTION[] = "FORMAT";
 
-    static constexpr std::string_view ICEBUG_DISK_PREFIX = "icebug-disk";
+    static constexpr std::string_view ICEBUG_DISK_FORMAT = "icebug-disk";
 
-    static bool isIceBugDiskStorage(const std::string& storage) {
-        return storage.starts_with(ICEBUG_DISK_PREFIX);
+    static bool isIceBugDiskFormat(const std::string& format) {
+        return format.find(ICEBUG_DISK_FORMAT) != std::string::npos;
     }
 };
 
--- a/src/include/storage/table/ice_disk_utils.h
+++ b/src/include/storage/table/ice_disk_utils.h
@@ -1,0 +1,133 @@
+#pragma once
+
+#include <algorithm>
+#include <cctype>
+#include <filesystem>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "common/constants.h"
+#include "common/exception/runtime.h"
+#include "common/file_system/virtual_file_system.h"
+#include "processor/operator/persistent/reader/parquet/parquet_reader.h"
+#include "storage/table/ice_disk_constants.h"
+
+namespace lbug {
+namespace storage {
+
+struct CSRFilePaths {
+    std::string indices;
+    std::string indptr;
+};
+
+class IceDiskUtils {
+public:
+    // Parses "icebug-disk", "icebug-disk:", or "icebug-disk:<path>" and returns the path
+    // component. Returns empty string for the first two forms (caller interprets as current dir)
+    static std::string getBasePath(const std::string& storage) {
+        if (!storage.starts_with(common::TableOptionConstants::ICEBUG_DISK_PREFIX)) {
+            return "";
+        }
+        std::string_view rest = std::string_view(storage).substr(
+            common::TableOptionConstants::ICEBUG_DISK_PREFIX.size());
+        // Strip the optional ':' separator.
+        if (!rest.empty() && rest[0] == ':') {
+            rest = rest.substr(1);
+        }
+        return std::string(rest); // empty means "current directory"
+    }
+
+    // Joins a base path with a filename. When base is empty the filename is returned
+    // as-is (i.e. relative to the current working directory)
+    static std::string joinPath(const std::string& base, const std::string& part) {
+        if (base.empty()) {
+            return part;
+        }
+        const char last = base.back();
+        if (last == '/' || last == '\\') {
+            return base + part;
+        }
+
+        return base + "/" + part;
+    }
+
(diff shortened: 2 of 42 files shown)


aheev:
dataset PR: https://github.com/LadybugDB/dataset/pull/1

aheev:
@adsharma could you PTAL? minimal test failure can be fixed by merging dataset [PR](https://github.com/LadybugDB/dataset/pull/1)

aheev:
Ran the tests added in the PR locally in CLI too. Also, `lbug -i schema.cypher` works fine

adsharma:
> minimal test failure can be fixed by merging dataset https://github.com/LadybugDB/dataset/pull/1

Could you add a commit to reset the submodule pointer? Tests should pass then. Also see my comment about making things work when schema.cypher is not in the same dir.

adsharma:
> (storage = 'icebug-disk:\<path-to-dir\>');

This doesn't leave room for saying `icebug-disk`  where the path is `s3://server.com/path-to-dir`, which the VFS supports (but not tested in CI because we don't have testable s3 urls).

How about:

```
(storage = '<any valid url>', format = 'icebug-disk')
```

The format could be guessed/defaulted based on the file extension in the URL.  This is consistent with how DuckDB does `EXPORT/IMPORT`.

I haven't reviewed the rest of the PR. But think it's directionally correct.

aheev:
> > (storage = 'icebug-disk:<path-to-dir>');
> 
> This doesn't leave room for saying `icebug-disk` where the path is `s3://server.com/path-to-dir`, which the VFS supports (but not tested in CI because we don't have testable s3 urls).
> 

The `path` is parsed out by strictly checking for `icebug-disk:` prefix

> How about:
> 
> ```
> (storage = '<any valid url>', format = 'icebug-disk')
> ```
> 

all in for separate dir path and format options

> The format could be guessed/defaulted based on the file extension in the URL. This is consistent with how DuckDB does `EXPORT/IMPORT`.
> 

storage is a `dir` path right? also defaulting to icebug-disk if the storage is a `dir` would block any future formats from using `storage` option

adsharma:
LGTM except for this item:

> all in for separate dir path and format options

Looks like you haven't implemented it yet? A test case for:

```
(storage = 's3://bucket/foo/bar', format = 'icebug-disk')
```

and we should be good to go. Also `format = 'bad'` should fail.

aheev:
> LGTM except for this item:
> 
> > all in for separate dir path and format options
> 
> Looks like you haven't implemented it yet? A test case for:
> 

I was waiting for your signal :upside_down_face:

aheev:
> (storage = 's3://bucket/foo/bar', format = 'icebug-disk')

this should fail right?

adsharma:
It's the main use case for icebug disk. I expect the `dir` on remote on object storage will be a common case. I couldn't tell if you agree with separating format and path logically by string splitting on `icebug-disk:` or a separate `format = x` option.

```
lbug> CREATE NODE TABLE city(id INT32, name STRING, population INT64, PRIMARY KEY(id)) WITH (storage = 'icebug-disk:s3://bucket/foo', format = 'icedisk');

...
Error: IO exception: Cannot open file s3:/bucket/foo/nodes_city.parquet: No such file or directory
```

adsharma:
> I was waiting for your signal

Missed this comment. I like the separate `format = 'x'` option, so `storage` is parsable with any URL parsing lib.

aheev:
> It's the main use case for icebug disk. I expect the `dir` on remote on object storage will be a common case. I couldn't tell if you agree with separating format and path logically by string splitting on `icebug-disk:` or a separate `format = x` option.
> 
> ```
> lbug> CREATE NODE TABLE city(id INT32, name STRING, population INT64, PRIMARY KEY(id)) WITH (storage = 'icebug-disk:s3://bucket/foo', format = 'icedisk');
> 
> ...
> Error: IO exception: Cannot open file s3:/bucket/foo/nodes_city.parquet: No such file or directory
> ```

separate format option

aheev:
new dataset PR: https://github.com/LadybugDB/dataset/pull/2

src/include/storage/table/ice_disk_utils.h
```diff
@@ -0,0 +1,106 @@
...
+struct CSRFilePaths {
+    std::string indices;
+    std::string indptr;
+};
+
+class IceDiskUtils {
+public:
+    // Joins a base path with a filename. When base is empty the filename is returned
+    // as-is (i.e. relative to the current working directory)
+    static std::string joinPath(const std::string& base, const std::string& part) {
```
adsharma, line 28: nit: private?

aheev:
@adsharma could you PTAL?

src/include/common/constants.h
```diff
@@ -86,6 +86,13 @@ struct StorageConstants {
 struct TableOptionConstants {
     static constexpr char REL_STORAGE_DIRECTION_OPTION[] = "STORAGE_DIRECTION";
     static constexpr char REL_STORAGE_OPTION[] = "STORAGE";
+    static constexpr char STORAGE_FORMAT_OPTION[] = "FORMAT";
+
+    static constexpr std::string_view ICEBUG_DISK_FORMAT = "icebug-disk";
+
+    static bool isIceBugDiskFormat(const std::string& format) {
+        return format.find(ICEBUG_DISK_FORMAT) != std::string::npos;
```
adsharma, line 94: Why not make it an exact match? Right now `(format 'not-icebug-format')` would pass this filter.

adsharma (approved):
Two minor issues - suggest making a small PR post merge.

Two major pre-existing issues (out of scope for this PR), but noting here for future reference:

  - High: src/storage/table/ice_disk_node_table.cpp:306 assigns nodeID.offset = rowIndex, where rowIndex is only the row’s position within the currently assigned  Parquet row group. With multiple row groups, every row group starts offsets at 0, producing duplicate node IDs and breaking joins/relationship traversals.
    This needs a row-group base offset from Parquet metadata added to the local row index.
  - High: src/storage/table/ice_disk_rel_table.cpp:311 drops relationships for later bound nodes in the same row group. When the scan sees a second boundOffset, it breaks, then scanInternalByRowGroups increments currentRowGroup, so the rest of that row group is never revisited. A query with a batch containing multiple source nodes can return only the first node’s relationships.

aheev:
> Two minor issues - suggest making a small PR post merge.
> 
> Two major pre-existing issues (out of scope for this PR), but noting here for future reference:
> 
> * High: src/storage/table/ice_disk_node_table.cpp:306 assigns nodeID.offset = rowIndex, where rowIndex is only the row’s position within the currently assigned  Parquet row group. With multiple row groups, every row group starts offsets at 0, producing duplicate node IDs and breaking joins/relationship traversals.
>   This needs a row-group base offset from Parquet metadata added to the local row index.
> * High: src/storage/table/ice_disk_rel_table.cpp:311 drops relationships for later bound nodes in the same row group. When the scan sees a second boundOffset, it breaks, then scanInternalByRowGroups increments currentRowGroup, so the rest of that row group is never revisited. A query with a batch containing multiple source nodes can return only the first node’s relationships.

Guess I have to resurrect https://github.com/LadybugDB/ladybug/pull/429 :sweat:

adsharma:
Because of the backward incompatible change in the catalog entry adding `storageFormat`, we can no longer open dbs created by 0.16.1. This will require a storage version bump and some compat glue.

adsharma:
Proposed fix: https://github.com/LadybugDB/ladybug/pull/489

aheev:
> Proposed fix: #489

Thanks!!

PR #486, opened 2026-05-12
Avoid syncing empty WAL transactions

### LINKED ISSUE #302: DuckDB is 2x faster than LadybugDB at non-batched inserts
Details:

https://github.com/LadybugDB/ladybug/discussions/301#discussioncomment-16264986

adsharma:
## Summary

Fixes: #302 

Use one fsync() per transaction instead of two. Previously we wrote two WAL records.

- Make LocalWAL log BEGIN lazily when the first real WAL record is added.
- Skip COMMIT logging for write transactions that produced no WAL records.
- Handle WAL size checks when no shared WAL writer/file exists.
- Add a regression for empty write transactions such as CALL THREADS=1.

## Root cause
Write transactions eagerly logged BEGIN at transaction start. If the transaction produced no WAL records, commit still appended COMMIT and flushed/synced an 18-byte BEGIN+COMMIT WAL entry when checksums were enabled. Making BEGIN lazy exposed an auto-checkpoint size check that assumed the shared WAL writer had been initialized; empty transactions now keep the writer uninitialized, so WAL::getFileSize() must treat that as size 0.

## Validation
- git diff --check
- make test-build-release
- build/release/test/transaction/transaction_test --gtest_filter=WalTest.EmptyWriteTransactionDoesNotCreateWAL

### DIFF
--- a/src/storage/wal/local_wal.cpp
+++ b/src/storage/wal/local_wal.cpp
@@ -18,14 +18,13 @@ LocalWAL::LocalWAL(MemoryManager& mm, bool enableChecksums)
       serializer(enableChecksums ? std::make_shared<ChecksumWriter>(inMemWriter, mm) :
                                    std::static_pointer_cast<Writer>(inMemWriter)) {}
 
-void LocalWAL::logBeginTransaction() {
-    BeginTransactionRecord walRecord;
-    addNewWALRecord(walRecord);
-}
-
 void LocalWAL::logCommit() {
+    std::unique_lock lck{mtx};
+    if (!hasLoggedBegin) {
+        return;
+    }
     CommitRecord walRecord;
-    addNewWALRecord(walRecord);
+    addNewWALRecordNoLock(walRecord);
 }
 
 void LocalWAL::logCreateCatalogEntryRecord(CatalogEntry* catalogEntry, bool isInternal) {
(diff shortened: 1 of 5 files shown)


adsharma:
Before:

```
➜  uv pip install ladybug
Resolved 1 package in 1ms
Installed 1 package in 2ms
 + ladybug==0.16.1
➜  uv run bench.py
Generating 1,000 rows...

--- Cypher CREATE (vanilla insert) ---
  Time : 8.164s
  Rows : 1,000
  Rate : 122 rows/s

--- Cypher MERGE (no conflicts) ---
  Time : 10.275s
  Rows : 1,000
  Rate : 97 rows/s

--- Cypher MERGE (50% conflicts) ---
  Time : 9.015s
  Rows : 1,974
  Rate : 111 rows/s
  ```
  
  After:
  
  ```
  uv run /tmp/t1/bench.py
Generating 1,000 rows...

--- Cypher CREATE (vanilla insert) ---
  Time : 4.390s
  Rows : 1,000
  Rate : 228 rows/s

--- Cypher MERGE (no conflicts) ---
  Time : 4.611s
  Rows : 1,000
  Rate : 217 rows/s

--- Cypher MERGE (50% conflicts) ---
  Time : 4.639s
  Rows : 1,981
  Rate : 216 rows/s
  ```

File: src/storage/wal/local_wal.cpp [hunk 1]
```diff
@@ -20,12 +20,18 @@ LocalWAL::LocalWAL(MemoryManager& mm, bool enableChecksums)
 
 void LocalWAL::logBeginTransaction() {
```
aheev: this func is no longer reqd I believe?

aheev (approved):
approved with a nit

PR #491, opened 2026-05-15
Fix ice-disk table scans

aheev:
- fixed nodeID offset in node table scan by calculating calc global row offset using parquet metadata
- fixed early break issue in rel table scans by refactoring ice-disk internal scan to full table based rather than rowGroup based
- enumized STORAGE_FORMAT

context: https://github.com/LadybugDB/ladybug/pull/476#pullrequestreview-4275243356

### DIFF
--- a/docs/icebug-disk.md
+++ b/docs/icebug-disk.md
@@ -25,8 +25,6 @@ CREATE REL TABLE livesin(FROM user TO city) WITH (storage = '<path-to-dir>', for
 
 File paths can be relative or absolute and are resolved as `<path-to-dir>/nodes_{tableName}.parquet` for node tables, and `<path-to-dir>/indices_{tableName}.parquet` and `<path-to-dir>/indptr_{tableName}.parquet` for relationship tables.
 
-Object-store URIs (e.g. `s3://bucket/path`, `https://host/path`) are supported as `storage` values and are passed through unchanged.
-
 If `storage` is omitted when `format = 'icebug-disk'` is set, files are resolved relative to the current working directory.
 
 Tables can also be created by manually running the above queries in the Ladybug CLI.
--- a/docs/icebug-disk.md
+++ b/docs/icebug-disk.md
@@ -25,6 +25,8 @@ CREATE REL TABLE livesin(FROM user TO city) WITH (storage = '<path-to-dir>', for
 
 File paths can be relative or absolute and are resolved as `<path-to-dir>/nodes_{tableName}.parquet` for node tables, and `<path-to-dir>/indices_{tableName}.parquet` and `<path-to-dir>/indptr_{tableName}.parquet` for relationship tables.
 
+Object-store URIs (e.g. `s3://bucket/path`, `https://host/path`) are also supported as `storage` values.
+
 If `storage` is omitted when `format = 'icebug-disk'` is set, files are resolved relative to the current working directory.
 
 Tables can also be created by manually running the above queries in the Ladybug CLI.
--- a/src/common/enums/storage_format.cpp
+++ b/src/common/enums/storage_format.cpp
@@ -1,0 +1,29 @@
+#include "common/enums/storage_format.h"
+
+#include "common/exception/binder.h"
+#include <format>
+
+namespace lbug {
+namespace common {
+
+StorageFormat StorageFormatUtils::fromString(const std::string& str) {
+    if (str == "icebug-disk") {
+        return StorageFormat::ICEBUG_DISK;
+    }
+    throw BinderException(
+        std::format("Unsupported storage format '{}'. Valid options are: icebug-disk.", str));
+}
+
+std::string StorageFormatUtils::toString(StorageFormat format) {
+    switch (format) {
+    case StorageFormat::NONE:
+        return "";
+    case StorageFormat::ICEBUG_DISK:
+        return "icebug-disk";
+    default:
+        return "";
+    }
+}
+
+} // namespace common
+} // namespace lbug
--- a/src/common/enums/storage_format.cpp
+++ b/src/common/enums/storage_format.cpp
@@ -14,16 +14,5 @@ StorageFormat StorageFormatUtils::fromString(const std::string& str) {
         std::format("Unsupported storage format '{}'. Valid options are: icebug-disk.", str));
 }
 
-std::string StorageFormatUtils::toString(StorageFormat format) {
-    switch (format) {
-    case StorageFormat::NONE:
-        return "";
-    case StorageFormat::ICEBUG_DISK:
-        return "icebug-disk";
-    default:
-        return "";
-    }
-}
-
 } // namespace common
 } // namespace lbug
--- a/src/include/catalog/catalog_entry/node_table_catalog_entry.h
+++ b/src/include/catalog/catalog_entry/node_table_catalog_entry.h
@@ -3,6 +3,7 @@
 #include <functional>
 #include <optional>
 
+#include "common/enums/storage_format.h"
 #include "function/table/table_function.h"
 #include "table_catalog_entry.h"
 
@@ -28,9 +29,9 @@ class LBUG_API NodeTableCatalogEntry final : public TableCatalogEntry {
 public:
     NodeTableCatalogEntry() = default;
     NodeTableCatalogEntry(std::string name, std::string primaryKeyName, std::string storage = "",
-        std::string storageFormat = "")
+        common::StorageFormat storageFormat = common::StorageFormat::NONE)
         : TableCatalogEntry{entryType_, std::move(name)}, primaryKeyName{std::move(primaryKeyName)},
-          storage{std::move(storage)}, storageFormat{std::move(storageFormat)} {}
+          storage{std::move(storage)}, storageFormat{storageFormat} {}
 
     // Constructor for foreign-backed tables
     NodeTableCatalogEntry(std::string name, std::string primaryKeyName,
@@ -57,7 +58,7 @@ public:
         return getProperty(primaryKeyName);
     }
     const std::string& getStorage() const { return storage; }
-    const std::string& getStorageFormat() const { return storageFormat; }
+    common::StorageFormat getStorageFormat() const { return storageFormat; }
     std::optional<function::TableFunction> getScanFunction() const override;
     const CreateBindDataFunc& getCreateBindDataFunc() const { return createBindDataFunc; }
     const std::string& getForeignDatabaseName() const { return foreignDatabaseName; }
@@ -84,7 +85,7 @@ private:
 private:
     std::string primaryKeyName;
     std::string storage;
-    std::string storageFormat;
+    common::StorageFormat storageFormat = common::StorageFormat::NONE;
     std::optional<function::TableFunction> scanFunction;
     CreateBindDataFunc createBindDataFunc; // Callback to create bind data
     std::string foreignDatabaseName;
--- a/src/processor/operator/scan/scan_rel_table.cpp
+++ b/src/processor/operator/scan/scan_rel_table.cpp
@@ -86,6 +86,8 @@ void ScanRelTable::initLocalStateInternal(ResultSet* resultSet, ExecutionContext
         scanState =
             std::make_unique<storage::IceDiskRelTableScanState>(*MemoryManager::Get(*clientContext),
                 boundNodeIDVector, outVectors, nbrNodeIDVector->state);
+        auto transaction = transaction::Transaction::Get(*context->clientContext);
+        iceDiskTable->initScanState(transaction, *scanState);
     } else if (foreignTable) {
         scanState =
             std::make_unique<storage::ForeignRelTableScanState>(*MemoryManager::Get(*clientContext),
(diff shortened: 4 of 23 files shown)


aheev:
@adsharma could you PTAL?

Re: duplicate boundNodes in unordered_map

```
Two cases:

 1. Source mode (MATCH (a:user)-[:follows]->(b) — direct node scan child): fetchNextBoundNodeBatch generates unique sequential offsets [nextOffset, nextOffset+N). No duplicates by construction.
 2. Non-source mode (multi-hop (a)-[r1]->(b)-[r2]->(c)):
 - r1's scan processes one source node a at a time (the break when boundOffset != activeBoundOffset)
 - So each call to r1.getNextTuple produces neighbors of exactly one a
 - A single source node's neighbor list has no duplicates in a well-formed CSR file
 - IceDisk node table emits 1 node per scan call. Even if it emits in a batch they would be distinct
 - Therefore r2's bound node vector always has distinct b values in each batch
 ```

**Comment by aheev** [comment]
dataset PR: https://github.com/LadybugDB/dataset/pull/3

**Comment by aheev** [comment]
@adsharma should we add a get_icebug_disk_supported_version CALL?

**Comment by adsharma** [comment]
We already have `db_version()` and `storage_version()`. Users can detect if icebug-disk is supported by trying `ATTACH`.

File: src/include/catalog/catalog_entry/node_table_catalog_entry.h [hunk 1]
```diff
@@ -84,7 +85,7 @@ class LBUG_API NodeTableCatalogEntry final : public TableCatalogEntry {
 private:
     std::string primaryKeyName;
     std::string storage;
-    std::string storageFormat;
+    common::StorageFormat storageFormat = common::StorageFormat::NONE;
```
adsharma, line 88: This is a backward incompatible change. But since we don't have any known prod usage of previous string storage format, I think we can survive without a storage version bump

src/storage/table/ice_disk_rel_table.cpp
```diff
@@ -187,133 +217,69 @@ void IceDiskRelTable::loadIndptrData(Transaction* transaction) const {
...
     auto activeBoundSelPos = INVALID_SEL;
     auto activeBoundOffset = INVALID_OFFSET;
     auto hasActiveBound = false;
 
-    // Calculate the starting global row index for the first row group
-    if (!rowGroupsToProcess.empty()) {
-        auto metadata = iceDiskScanState.indicesReader->getMetadata();
-        for (uint64_t rgIdx = 0; rgIdx < rowGroupsToProcess[0]; ++rgIdx) {
-            currentGlobalRowIdx += metadata->row_groups[rgIdx].num_rows;
+    while (totalRowsCollected < maxRowsPerCall) {
```
adsharma, line 243: #### 1. Infinite loop in scanInternal — ice_disk_rel_table.cpp:239–340

 When the scanner encounters two different bound nodes within the same cached batch and hasn't yet filled the output vector (totalRowsCollected <
 maxRowsPerCall), the outer while loop becomes an infinite loop.

 Trace:
 1. Scan rows for bound node A, collect 3 rows, totalRowsCollected = 3
 2. Hit row belonging to bound node B → break the inner for loop (does not increment currentLocalRowIdx)
 3. Outer while (3 < 2048) is still true; batch reload check fails (not exhausted)
 4. Inner for restarts at B's row; hasActiveBound = true, activeBoundOffset = A → immediate break
 5. Repeat forever

 This fires in source-mode scans (e.g. MATCH (a:user)-[:follows]->(b:user)) where fetchNextBoundNodeBatch packs multiple bound nodes into one boundNodeOffsets
 map.

 Fix: after the inner for exits due to a different bound node, break the outer loop too:

 ```cpp
   bool differentBoundEncountered = false;
   for (...) {
       ...
       } else if (boundOffset != activeBoundOffset) {
           differentBoundEncountered = true;
           break;
       }
       ...
   }
   if (differentBoundEncountered) break; // exit outer while, return what we have
 ```

src/storage/table/ice_disk_rel_table.cpp
```diff
@@ -66,28 +85,6 @@ void IceDiskRelTable::initScanState(Transaction* transaction, TableScanState& sc
...
-    }
-    if (!indptrFilePath.empty() && !iceDiskScanState.indptrReader) {
-        std::vector<bool> columnSkips; // Read all columns
-        auto context = transaction->getClientContext();
-        iceDiskScanState.indptrReader =
-            std::make_unique<ParquetReader>(indptrFilePath, columnSkips, context);
-    }
-
-    // Load shared indptr data - thread-safe to read
-    if (!indptrFilePath.empty()) {
```
adsharma, line 87: This guard was significant?

src/storage/table/ice_disk_node_table.cpp
```diff
@@ -307,11 +307,21 @@ bool IceDiskNodeTable::scanInternal(Transaction* transaction, TableScanState& sc
         }
     }
 
+    // calc current global row index based on assigned row group and local row index within that
+    // group
+    auto metadata = iceDiskScanState.parquetReader->getMetadata();
+    offset_t startOffset = 0;
```
adsharma, line 313: startOffset for a given nodeGroupIdx is constant?

src/processor/operator/scan/scan_rel_table.cpp
```diff
@@ -86,6 +86,8 @@ void ScanRelTable::initLocalStateInternal(ResultSet* resultSet, ExecutionContext
         scanState =
             std::make_unique<storage::IceDiskRelTableScanState>(*MemoryManager::Get(*clientContext),
                 boundNodeIDVector, outVectors, nbrNodeIDVector->state);
+        auto transaction = transaction::Transaction::Get(*context->clientContext);
+        iceDiskTable->initScanState(transaction, *scanState);
```
adsharma: One of my agents flagged this as potentially doing unnecessary work. Full comment:

  - [P2] Avoid initializing IceDisk rel scans before a bound-node batch exists — src/processor/operator/scan/scan_rel_table.cpp:89
    initLocalStateInternal now calls iceDiskTable->initScanState(...) immediately after constructing the scan state, before tableInfo.initScanState(...) and
  before the first child/source batch has populated nodeIDVector. The cached bound selection is empty at that point, but the first table->scan(...) still walks
  the whole IceDisk indices parquet looking for matches in an empty boundNodeOffsets map, then discards the result and reinitializes after fetching the first
  real batch. On large IceDisk rel tables this makes every rel scan pay an unnecessary full-file pass before producing any rows. The IceDisk init should happen
  only when a real bound-node batch is ready, as the later calls at lines 129/164 already do.

docs/icebug-disk.md
```diff
@@ -25,8 +25,6 @@ CREATE REL TABLE livesin(FROM user TO city) WITH (storage = '<path-to-dir>', for
 
 File paths can be relative or absolute and are resolved as `<path-to-dir>/nodes_{tableName}.parquet` for node tables, and `<path-to-dir>/indices_{tableName}.parquet` and `<path-to-dir>/indptr_{tableName}.parquet` for relationship tables.
 
-Object-store URIs (e.g. `s3://bucket/path`, `https://host/path`) are supported as `storage` values and are passed through unchanged.
```
adsharma, line 28: Why drop this? Should still hold afterwards?

src/common/enums/storage_format.cpp
```diff
@@ -0,0 +1,29 @@
...
+}
+
+std::string StorageFormatUtils::toString(StorageFormat format) {
+    switch (format) {
+    case StorageFormat::NONE:
+        return "";
+    case StorageFormat::ICEBUG_DISK:
+        return "icebug-disk";
+    default:
+        return "";
```
adsharma: This should throw.

src/storage/table/ice_disk_rel_table.cpp
```diff
@@ -42,6 +42,25 @@ void IceDiskRelTableScanState::setToTable(const Transaction* transaction, Table*
     // IceDiskRelTable does not support local storage, so we skip the local table initialization
 }
 
+void IceDiskRelTableScanState::reloadCachedBatchData(Transaction* transaction) {
+    auto context = transaction->getClientContext();
+
+    // Create DataChunk matching the indices parquet file schema
+    auto numIndicesColumns = indicesReader->getNumColumns();
+    cachedBatchData = std::make_unique<DataChunk>(numIndicesColumns);
```
adsharma, line 50: Can these allocations be done once on reset() and reused?

src/storage/table/ice_disk_rel_table.cpp
```diff
@@ -42,6 +42,25 @@ void IceDiskRelTableScanState::setToTable(const Transaction* transaction, Table*
...
+
+    // Insert value vectors for all columns in the parquet file
+    auto memoryManager = MemoryManager::Get(*context);
+    for (uint32_t colIdx = 0; colIdx < numIndicesColumns; ++colIdx) {
+        const auto& columnTypeRef = indicesReader->getColumnType(colIdx);
+        auto columnType = columnTypeRef.copy();
+        auto vector = std::make_shared<ValueVector>(std::move(columnType), memoryManager);
```
adsharma, line 57: Ditto

src/include/catalog/catalog_entry/node_table_catalog_entry.h
adsharma: Actually, this fails to open a version 41 database:

```
./build/release/tools/shell/lbug -r /tmp/test1-save.db
Terminate called after throwing an instance of std::bad_alloc: std::bad_alloc
<while deserializing catalog entry>
```

Either revert or implement in-place upgrade.

src/storage/table/ice_disk_rel_table.cpp
aheev: indptr and indices path validation is done during table creation phase

src/include/catalog/catalog_entry/node_table_catalog_entry.h
aheev: is it created with the bin of this branch?

src/processor/operator/scan/scan_rel_table.cpp
aheev: Nice find. I will check for boundOffsets.empty first in scanInternal

src/storage/table/ice_disk_node_table.cpp
aheev: startOffset for a nodeGroupIdx(rowGroup) is calc just below. We can avoid this repeated calc by populating startOffsets in initGlobalStateInternal. I will add it in refactor post release. Keeping changes minimal right now

docs/icebug-disk.md
aheev: we can add it later once we support URIs

src/storage/table/ice_disk_rel_table.cpp
aheev: same as above

src/storage/table/ice_disk_rel_table.cpp
aheev: DataChunk doesn't offer a `reset` out of the box. All it offers is resetAuxiliaryBuffer. We need to manually reset `state` in DataChunk and other state objects in ValueVectors which requires tinkering with ParquetReader and/or ValueVector. Maybe refactor it later?

aheev:
new dataset PR: https://github.com/LadybugDB/dataset/pull/4

src/include/catalog/catalog_entry/node_table_catalog_entry.h
adsharma: dbs created from 0.16.0 and from a bin from main (without this PR) will have the same outcome. 

The way kuzu dealt with it was to ask users to export/import db after every release. Since we went several releases with version 40 and backward compat, this would be a good contract to maintain going forward.

aheev: loads up fine for me with both 0.16.1 and 0.15.3. I haven't removed `upgradeLegacyStorageFormat` from my previous PR

docs/icebug-disk.md
adsharma: The VFS has supported URIs for a long time. We haven't touched that code, but we haven't verified them either (because all the s3/httpfs assets were not accessible after kuzu archiving).

I'm sending a small PR to thread the shell through VFS and change path mangling to restore URIs. Fortunately most of it seems to be intact and functional.

aheev: are you saying parquet_reader code would work with URIs too?

adsharma: https://github.com/LadybugDB/ladybug/pull/493

adsharma: Yes - tested via tooling in #493. Same result local file or http:// url.

s3:// should work too once we figure out ordering of lbug -i <init url> processing vs cred handling.

src/include/catalog/catalog_entry/node_table_catalog_entry.h
adsharma: Handled correctly in:

src/catalog/catalog_entry/rel_group_catalog_entry.cpp:150

adsharma:
Nice improvements! Ok to handle minor unresolved comments separately.

PR #492, opened 2026-05-16
Implement Adaptive Radix Tree (ART) Indexes

adsharma:
Details in the included docs.

```
du -sh /tmp/test1*.db
634M    /tmp/test1-hash.db
512M    /tmp/test1-noindex.db
604M    /tmp/test1-art.db
```

```
➜  ladybug git:(art_index) ✗ ./build/release/tools/shell/lbug -r /tmp/test1-hash.db
Opening the database at path: /tmp/test1-hash.db in read-only mode.
Enter ":help" for usage hints.
lbug> CALL show_indexes() return *;
┌────────────┬──────────────┬────────────┬────────────────┬──────────────────┬──────────────────────────────────────────────────────────────┐
│ table_name │ index_name   │ index_type │ property_names │ extension_loaded │ index_definition                                             │
│ STRING     │ STRING       │ STRING     │ STRING[]       │ BOOL             │ STRING                                                       │
├────────────┼──────────────┼────────────┼────────────────┼──────────────────┼──────────────────────────────────────────────────────────────┤
│ User       │ user_hash_pk │ HASH       │ [id]           │ True             │ CREATE HASH INDEX `user_hash_pk` FOR (n:`User`) ON (n.`id`); │
└────────────┴──────────────┴────────────┴────────────────┴──────────────────┴──────────────────────────────────────────────────────────────┘
(1 tuple)
(6 columns)
Time: 5.56ms (compiling), 1.69ms (executing)
lbug>
➜  ladybug git:(art_index) ✗ ./build/release/tools/shell/lbug -r /tmp/test1-art.db
Opening the database at path: /tmp/test1.db in read-only mode.
Enter ":help" for usage hints.
lbug> CALL show_indexes() return *;
┌────────────┬───────────────┬────────────┬────────────────┬──────────────────┬──────────────────────────────────────────────────────────────┐
│ table_name │ index_name    │ index_type │ property_names │ extension_loaded │ index_definition                                             │
│ STRING     │ STRING        │ STRING     │ STRING[]       │ BOOL             │ STRING                                                       │
├────────────┼───────────────┼────────────┼────────────────┼──────────────────┼──────────────────────────────────────────────────────────────┤
│ User       │ idx_person_pk │ ART        │ [id]           │ True             │ CREATE ART INDEX `idx_person_pk` FOR (n:`User`) ON (n.`id`); │
└────────────┴───────────────┴────────────┴────────────────┴──────────────────┴──────────────────────────────────────────────────────────────┘
(1 tuple)
(6 columns)
Time: 2.08ms (compiling), 0.21ms (executing)
lbug>
➜  ladybug git:(art_index) ✗ ./build/release/tools/shell/lbug -r /tmp/test1-noindex.db
Opening the database at path: /tmp/test1-save.db in read-only mode.
Enter ":help" for usage hints.
lbug> CALL show_indexes() return *;
┌────────────┬────────────┬────────────┬────────────────┬──────────────────┬──────────────────┐
│ table_name │ index_name │ index_type │ property_names │ extension_loaded │ index_definition │
│ STRING     │ STRING     │ STRING     │ STRING[]       │ BOOL             │ STRING           │
├────────────┼────────────┼────────────┼────────────────┼──────────────────┼──────────────────┤
└────────────┴────────────┴────────────┴────────────────┴──────────────────┴──────────────────┘
(0 tuples)
(6 columns)
Time: 1.50ms (compiling), 0.16ms (executing)
```

### DIFF
--- a/docs/art_index.md
+++ b/docs/art_index.md
@@ -110,3 +110,10 @@ serialized into the existing index catalog storage-info buffer.
 On reload, the ART index deserializes the key/offset pairs and rebuilds the
 in-memory tree. Built-in `ART` catalog entries are marked as loaded in the same
 way as built-in `HASH` entries so inserts can continue after reload.
+
+The persisted ART is not currently page-backed or demand loaded. Opening a
+database with an ART index materializes the full in-memory tree by replaying all
+stored key/offset entries. Memory use is therefore O(number of indexed keys),
+not O(number of queried index pages or blocks). A production disk-backed ART
+would need buffer-managed index pages and lazy traversal instead of loading the
+entire index at startup.
--- a/docs/art_index.md
+++ b/docs/art_index.md
@@ -1,0 +1,84 @@
+# ART Index
+
+Ladybug supports a built-in adaptive radix tree index for node primary keys:
+
+```cypher
+CREATE ART INDEX person_pk FOR (p:Person) ON (p.ID);
+```
+
+This is implemented by `storage::ArtPrimaryKeyIndex`, registered as the built-in
+`ART` index type. It is parallel to the existing `HASH` primary-key index in the
+DDL and table lookup paths, but it uses a radix tree keyed by encoded primary-key
+bytes instead of a hash table.
+
+## Scope
+
+The ART index currently supports Ladybug's primary-key index workflow:
+
+- building the index from an existing node table during `CREATE ART INDEX`
+- checking uniqueness for committed inserts
+- primary-key lookup through `NodeTable::lookupPK`
+- rollback cleanup for failed insert batches
+- checkpoint and reload
+
+It does not currently provide secondary-index scans, range scans, prefix
+compression, duplicate row-id leaves, or DuckDB's parallel local-ART merge
+builder.
+
+## Key Encoding
+
+`ArtKey::encode` converts a primary-key value into a byte vector. Null keys are
+not valid primary keys and are rejected during insert.
+
+Strings are encoded using the same escaping idea as DuckDB's ART keys:
+
+- bytes `0x00` and `0x01` are prefixed with `0x01`
+- a trailing `0x00` terminator is appended
+
+Fixed-width numeric values are encoded as their in-memory bytes. This is enough
+for the current equality-only primary-key lookup path. If ART is later extended
+to ordered scans, numeric encodings will need to become order-preserving.
+
+## Tree Shape
+
+Each tree node may store an optional node-table offset for a complete key and a
+set of byte-labeled children. Children use adaptive layouts:
+
+- `Node4`: up to 4 children
+- `Node16`: up to 16 children
+- `Node48`: up to 48 children, with a 256-byte indirection table
+- `Node256`: direct child pointer array for all byte values
+
+Nodes grow as children are inserted. Lookups walk one encoded key byte at a time
+from the root to a terminal node and then verify visibility before returning the
+stored offset.
+
+## Inserts And Uniqueness
+
+`ArtPrimaryKeyIndex::commitInsert` is called through the generic index commit
+path. For each row, it encodes the primary-key value and inserts the key into
+the radix tree with the row offset from the node ID vector.
+
+If the terminal node already has a visible offset, insertion fails with the same
+duplicate-primary-key error used by the hash index. If the existing offset is not
+visible to the caller, the stored offset may be replaced.
+
+## Lookup Path
+
+`NodeTable::lookupPK` now asks the loaded primary-key index to perform a generic
+`lookupPrimaryKey` operation. The hash index implements this by delegating to
+its existing lookup code. The ART index encodes the lookup key, walks the tree,
+and returns the offset only if the supplied visibility function accepts it.
+
+This keeps the planner and table lookup path independent of whether the physical
+primary-key index is `HASH` or `ART`.
+
+## Checkpoint And Reload
+
+During checkpoint, the ART index collects all terminal key/offset pairs from the
+tree and stores them in `ArtPrimaryKeyIndexStorageInfo`. The storage info is
+serialized into the existing index catalog storage-info buffer.
+
+On reload, the ART index deserializes the key/offset pairs and rebuilds the
+in-memory tree. Built-in `ART` catalog entries are marked as loaded in the same
+way as built-in `HASH` entries so inserts can continue after reload.
--- a/docs/art_index.md
+++ b/docs/art_index.md
@@ -18,12 +18,16 @@ The ART index currently supports Ladybug's primary-key index workflow:
 - building the index from an existing node table during `CREATE ART INDEX`
 - checking uniqueness for committed inserts
 - primary-key lookup through `NodeTable::lookupPK`
+- bounded primary-key range scans for constant inequality predicates
 - rollback cleanup for failed insert batches
 - checkpoint and reload
 
-It does not currently provide secondary-index scans, range scans, prefix
-compression, duplicate row-id leaves, or DuckDB's parallel local-ART merge
-builder.
+It does not currently provide secondary indexes, prefix compression, duplicate
+row-id leaves, or DuckDB's parallel local-ART merge builder. Range scans are
+used only for single-table constant predicates over an ART-backed primary key.
+The optimizer keeps the original range predicates as residual filters, so the
+range scan is an access-path optimization rather than a replacement for general
+predicate evaluation.
 
 ## Key Encoding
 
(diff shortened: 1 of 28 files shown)


adsharma:
```
random_lookup_bench.py --backend pybind --literal --lookups 500 --warmup 50


  Literal benchmark cross-check, 500 lookups:

  base  1996.8/s  avg=0.501ms p95=0.615ms
  hash  1227.3/s  avg=0.815ms p95=1.802ms
  art  12029.0/s  avg=0.083ms p95=0.095ms

Without --literal  
  
  base  1405.7/s avg=0.711ms p95=0.842ms
  hash  1350.4/s avg=0.740ms p95=0.959ms
  art   1424.0/s avg=0.702ms p95=0.817ms

```

 So with literal constants, ART is clearly being used and is much faster. The prepared benchmark is still mostly tied, which suggests the prepared/parameter path or Python execution overhead is masking the index difference. Hash being slower than base in the literal run is also notable.
 
 https://gist.github.com/adsharma/83a1b7c9c320d829e349135d396ab4d3

adsharma:
ART index also enables range scans. Previously ladybug/kuzu didn't support range scans via indexes.

src/include/storage/index/art_index.h
```diff
@@ -0,0 +1,136 @@
...
+        bool upperInclusive, common::idx_t maxResults, std::vector<common::offset_t>& results,
+        visible_func isVisible) override;
+    void discardPrimaryKey(common::ValueVector* keyVector) override;
+
+    void checkpoint(main::ClientContext*, PageAllocator&) override;
+
+    static std::unique_ptr<Index> load(main::ClientContext* context, StorageManager* storageManager,
+        IndexInfo indexInfo, std::span<uint8_t> storageInfoBuffer);
+
+    static IndexType getIndexType() {
```
aheev, line 93: can teh definition be moved to source file?

src/optimizer/filter_push_down_optimizer.cpp
```diff
@@ -321,6 +342,51 @@ std::shared_ptr<Expression> PredicateSet::popNodePKEqualityComparison(const Expr
...
+            comparisonType = ExpressionTypeUtil::reverseComparisonDirection(comparisonType);
+        } else {
+            continue;
+        }
+        if (!isConstantExpression(bound)) {
+            continue;
+        }
+        switch (comparisonType) {
+        case ExpressionType::GREATER_THAN:
+            result.lowerBound = bound;
```
aheev, line 377: what if the user passes multiple same comparisonType predicates? like `id >= 5 AND id > 10`

src/storage/index/art_index.cpp
```diff
@@ -0,0 +1,460 @@
...
+    const ArtKey* lowerBound, bool lowerInclusive, const ArtKey* upperBound, bool upperInclusive,
+    idx_t maxResults, std::vector<offset_t>& results, visible_func isVisible) const {
+    if (results.size() >= maxResults) {
+        return;
+    }
+    if (node.offset.has_value() && satisfiesLowerBound(key, lowerBound, lowerInclusive) &&
+        satisfiesUpperBound(key, upperBound, upperInclusive) && isVisible(node.offset.value())) {
+        results.push_back(node.offset.value());
+    }
+    for (auto byte = 0u; byte <= std::numeric_limits<uint8_t>::max(); ++byte) {
```
aheev: from copilot:

 src/storage/index/art_index.cpp — collectRange

```cpp
   for (uint16_t byte = 0; byte <= 255; byte++) {
       Node* child = getChild(node, byte);  // O(n) scan for NODE4/NODE16
       if (child) { ... }
   }
```

  For NODE4/NODE16 getChild does a linear scan; the outer loop calls it 256 times → up to 1,024 comparisons per node. Since smallChildren and keys are already sorted, you could iterate existing children in order and stop/skip outside the range, reducing to O(numChildren) per node.

src/include/storage/index/art_index.h
```diff
@@ -0,0 +1,136 @@
...
+        visible_func isVisible) override;
+    bool scanPrimaryKeyRange(common::ValueVector* lowerBoundVector, uint64_t lowerBoundPos,
+        bool lowerInclusive, common::ValueVector* upperBoundVector, uint64_t upperBoundPos,
+        bool upperInclusive, common::idx_t maxResults, std::vector<common::offset_t>& results,
+        visible_func isVisible) override;
+    void discardPrimaryKey(common::ValueVector* keyVector) override;
+
+    void checkpoint(main::ClientContext*, PageAllocator&) override;
+
+    static std::unique_ptr<Index> load(main::ClientContext* context, StorageManager* storageManager,
```
aheev: should this be exported same as `HashIndex::load()`?

src/optimizer/filter_push_down_optimizer.cpp
aheev: Also, shouldn't we pop the predicates(incl in the result) from the predicateSet?

docs/art_index.md
```diff
@@ -0,0 +1,106 @@
...
+- primary-key lookup through `NodeTable::lookupPK`
+- bounded primary-key range scans for constant inequality predicates
+- rollback cleanup for failed insert batches
+- checkpoint and reload
+
+It does not currently provide secondary indexes, prefix compression, duplicate
+row-id leaves, or DuckDB's parallel local-ART merge builder. Range scans are
+used only for single-table constant predicates over an ART-backed primary key.
+The optimizer keeps the original range predicates as residual filters, so the
+range scan is an access-path optimization rather than a replacement for general
```
aheev: what's the motive behind this design choice?

src/include/storage/index/art_index.h
```diff
@@ -0,0 +1,136 @@
...
+private:
+    struct Node {
+        static constexpr uint8_t EMPTY_MARKER = UINT8_MAX;
+
+        enum class Kind : uint8_t { NODE4 = 0, NODE16 = 1, NODE48 = 2, NODE256 = 3 };
+
+        std::optional<common::offset_t> offset;
+        Kind kind = Kind::NODE4;
+        uint16_t count = 0;
+        std::array<uint8_t, 16> keys{};
```
aheev: should we have separate structs for different nodeTypes to save on memory?

src/storage/index/art_index.cpp
```diff
@@ -0,0 +1,460 @@
...
+    }
+    auto* node = &root;
+    for (const auto byte : key.getBytes()) {
+        auto* child = node->getChild(byte);
+        if (child == nullptr) {
+            return;
+        }
+        node = child;
+    }
+    node->offset.reset();
```
aheev: shouldn't we delete the node too?

aheev: Also what about shrinking the tree?

aheev (changes requested):
queries after COPY FROM are failing

```
-CASE ArtIndexCopyFrom
-STATEMENT CALL enable_default_hash_index=false;
---- ok
-STATEMENT CREATE NODE TABLE art_copy_person (ID INT64, name STRING, PRIMARY KEY(ID));
---- 1
Table art_copy_person has been created.
-STATEMENT CREATE ART INDEX art_copy_person_pk FOR (p:art_copy_person) ON (p.ID);
---- 1
Index art_copy_person_pk has been created.
-STATEMENT COPY art_copy_person FROM "${LBUG_ROOT_DIRECTORY}/dataset/art-index-test/person.csv";
---- 1
4 tuples have been copied to the art_copy_person table.
-STATEMENT MATCH (p:art_copy_person) WHERE p.ID = 2 RETURN p.name;
---- 1
Grace
-STATEMENT MATCH (p:art_copy_person) WHERE p.ID >= 1 AND p.ID <= 10 RETURN p.ID, p.name ORDER BY p.ID;
---- 3
1|Ada
2|Grace
10|Barbara
-STATEMENT MATCH (p:art_copy_person) WHERE p.ID < 2 RETURN p.ID, p.name ORDER BY p.ID;
---- 2
-1|Edsger
1|Ada
-STATEMENT CALL enable_default_hash_index=true;
---- ok
```

[person.csv](https://github.com/user-attachments/files/27857346/person.csv)

We should also look at thread safety

docs/art_index.md
adsharma: To keep the scope of the diff small. Otherwise this will spiral into implementing range scan infra in ladybug which didn't exist before the PR

adsharma:
Key changes:

  - ART index creation now validates the physical contract: built-in primary-key index, exactly one indexed column, exactly one key type, supported scalar key
    type.
  - Range pushdown now only uses ART range scan for validated simple shapes: at most one constant lower bound and one constant upper bound. Duplicate same-side
    bounds or complex/non-constant predicates stay as normal filters.
  - Consumed valid range predicates are now removed from the residual predicate set.
  - ART public lookup/scan/insert/discard/checkpoint paths now take a coarse mutex for thread safety.
  - Moved initInsertState out of the header and marked ArtPrimaryKeyIndex::load with LBUG_API.
  - Reworked range traversal to iterate actual children instead of probing all 256 byte values through getChild.
  - Rollback/discard cleanup now prunes empty child nodes; it does not shrink node kind layouts, which I’d keep as a separate memory-layout refactor.
  - Updated docs/art_index.md with unsupported shapes and optimizer limitations.
  - Added e2e coverage for unsupported ART creation, duplicate/complex range predicate behavior, and kept the COPY regression.

aheev (approved):

adsharma:
Need refactor the pathological unique_ptr<> deallocation pattern in this PR. Eventually we should use a more efficient purpose designed memory allocator for the nodes. But for now, I'll try to avoid the long pause on shell exit.

PR #507, opened 2026-05-22
Register ADBC extension

adsharma:
This will allow us to query more columnar data sources that speak ADBC

More info about [adbc-drivers.](https://arrow.apache.org/adbc/current/index.html)

## Runtime testing status:

  - LOAD EXTENSION .../libadbc.lbug_extension works.
  - ATTACH ... (DBTYPE ADBC, DRIVER=..., TABLES='person') works.
  - Scanning currently fails at ArrowArrayStream.get_next from DuckDB ADBC with:

  Invalid Input Error: Attempting to execute an unsuccessful or closed pending query result

That points to a DuckDB ADBC direct-driver lifetime/manager expectation, not a compile issue. DuckDB’s docs describe using the ADBC driver-manager-style setup  with driver and entrypoint options, while the direct initialized vtable rejects those options. Source: https://duckdb.org/docs/current/clients/adbc.html

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -254,6 +254,20 @@ jobs:
           large-packages: true
           swap-storage: true
 
+      - name: Setup pixi
+        uses: prefix-dev/setup-pixi@v0.9.4
+        with:
+          pixi-version: v0.68.1
+          cache: false
+          activate-environment: true
+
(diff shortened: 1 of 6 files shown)


adsharma:
Another way to do ADBC: https://duckdb.org/community_extensions/extensions/adbc_scanner

PR #508, opened 2026-05-22
Fix/windows storage version check

adsharma:
- Fix nightly test failures related to windows file locking
- Today's arrow updates triggered another windows symptom: lbug.lib > 4GB + 400kb in size. This triggers build errors.
- Fixed by switching to shared instead of static lbug library.


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/26268762851

PR #509, opened 2026-05-22
icebug-disk: relax metadata checks

adsharma:
```
# link httpfs statically into lbug
```

works now. But two problems remain:

- Many roundtrips opening all the files at startup. Perhaps some of it can be delayed
- queries are slow

```
lbug> match (a)-[b:followedby]->(c) return a.*, b.*, c.*;
...
(40 tuples, 20 shown)
(52 columns, 10 shown)
Time: 630.63ms (compiling), 16320.14ms (executing)
```


aheev:
LGTM

Issue #170, opened 2026-02-07
Bug: Merge node with gen_random_uuid() primary key returns wrong primary key on match


andreasdamm-shure:
### Ladybug version

0.14.1

### What operating system are you using?

Windows 11

### What happened?

When running a merge clause from an unwind statement for a node that uses gen_random_uuid() as its default primary key, different primary keys a returned between create and match case. 

### Are there known steps to reproduce?

Using the Kuzu CLI run the following queries

```cypher
create node table A(id string default gen_random_uuid(), stuff int8, primary key (id));

unwind [1, 1] as i
      merge (a:A { stuff: i })
      return a.id;
```

The output should look like
```
┌──────────────────────────────────────┐
│ a.id                                 │
│ STRING                               │
├──────────────────────────────────────┤
│ 39d946c2-01ee-466f-a9b2-98f2e6bb54fa │
│ 39d946c2-01ee-466f-a9b2-98f2e6bb54fa │
└──────────────────────────────────────┘
```

but instead on the first run it looks like

```
┌──────────────────────────────────────┐
│ a.id                                 │
│ STRING                               │
├──────────────────────────────────────┤
│ 39d946c2-01ee-466f-a9b2-98f2e6bb54fa │
│ e25ac114-5764-45c7-b6c7-1db7a407c8c9 │
└──────────────────────────────────────┘
```

i.e. two different ids are returned when they should have been the same.

adsharma:
Good:

```
lbug> explain
            merge (a:A { stuff: 1 })
            return a.id;
┌─────────────────────────────┐
│┌───────────────────────────┐│
││       Physical Plan       ││
│└───────────────────────────┘│
└─────────────────────────────┘
┌─────────────────────────────┐
│     RESULT_COLLECTOR[7]     │
│   -----------------------   │
│      Expressions: a.id      │
│   -----------------------   │
│     NumOutputTuples: 0      │
│   -----------------------   │
│   ExecutionTime: 0.000000   │
└──────────────┬──────────────┘
┌──────────────┴──────────────┐
│        PROJECTION[6]        │
│   -----------------------   │
│      Expressions: a.id      │
│   -----------------------   │
│     NumOutputTuples: 0      │
│   -----------------------   │
│   ExecutionTime: 0.000000   │
└──────────────┬──────────────┘
┌──────────────┴──────────────┐
│          MERGE[5]           │
│   -----------------------   │
│        Pattern: a.id        │
│           a.stuff           │
[...]

PR #510, opened 2026-05-22
fix unwind merge relation issue

ericyuanhui:
src/optimizer/unwind_dedup_optimizer.cpp:45
  Blocking correctness concern: the dedup key is inferred as “all expressions in the probe schema that depend on the unwind variable”, not the actual `MERGE` identity. That can drop legitimate rows when one input row expands to multiple matched nodes.

  Example shape:

```
  UNWIND [{age: 1, dept_id: 'd'}] AS row
  WITH row.age AS age, row.dept_id AS dept_id
  MATCH (p:Person {age: age}), (d:Department {id: dept_id})
  MERGE (p)-[:R]->(d)
```

 If two Person nodes have `age = 1`, this should create two relationships.  The buggy UNWIND_DEDUP operator creates only one relationship because of the dedup key structure.
  
 Fixed by placing the dedup where the actual merge identity is available.

### DIFF
--- a/src/include/optimizer/unwind_dedup_optimizer.h
+++ b/src/include/optimizer/unwind_dedup_optimizer.h
@@ -8,11 +8,11 @@ namespace optimizer {
 
 /* When UNWIND is followed by MERGE, duplicate values in the UNWIND list can cause
  * issues because the MERGE's HASH_JOIN doesn't see newly created nodes within the same batch.
- * This optimizer inserts a UNWIND_DEDUP operator to deduplicate UNWIND output before MERGE.
+ * This optimizer inserts a UNWIND_DEDUP operator to deduplicate MERGE input rows.
  *
  * E.g., UNWIND [1, 1, 2] AS x MERGE (a:A {val: x})
  * Before: UNWIND -> HASH_JOIN (for optional match) -> MERGE
- * After:  UNWIND -> UNWIND_DEDUP -> HASH_JOIN -> MERGE
+ * After:  UNWIND -> HASH_JOIN (for optional match) -> UNWIND_DEDUP -> MERGE
  */
 class UnwindDedupOptimizer : public LogicalOperatorVisitor {
 public:
@@ -20,10 +20,12 @@ public:
 
 private:
     std::shared_ptr<planner::LogicalOperator> visitOperator(
-        const std::shared_ptr<planner::LogicalOperator>& op);
+        const std::shared_ptr<planner::LogicalOperator>& op, bool isRoot = false);
 
     std::shared_ptr<planner::LogicalOperator> visitMergeReplace(
         std::shared_ptr<planner::LogicalOperator> op) override;
+
+    bool canRewriteCurrentMerge = false;
 };
 
 } // namespace optimizer
--- a/src/include/planner/operator/logical_unwind_deduplicate.h
+++ b/src/include/planner/operator/logical_unwind_deduplicate.h
@@ -10,22 +10,22 @@ class LogicalUnwindDeduplicate final : public LogicalOperator {
 
 public:
     LogicalUnwindDeduplicate(std::shared_ptr<LogicalOperator> child,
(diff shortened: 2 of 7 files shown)


adsharma:
A more general version of the fix is possible, but I'll defer it to a future PR. Some notes here for future reference:

 It currently only fires for this shape:

```
  MERGE
    HASH_JOIN
      probe side containing UNWIND somewhere
      build side / match side
```

  The “more correct” part is that it no longer guesses keys from the UNWIND projection. It waits until MERGE’s direct child has produced the actual LogicalMerge::getKeys(), then dedups there:

```
  MERGE
    UNWIND_DEDUP(keys = merge keys)
      HASH_JOIN
        ...
```

  So if the merge identity is available at MERGE’s child, it works even if the UNWIND is nested under projections/flatten/etc. inside the join probe.

  But your deeper-plan concern is valid. If the plan is more like:

```
  MERGE
    OP_LATER
      OP_MORE
        HASH_JOIN
          ...
            UNWIND
```

  then this optimizer currently won’t fire, because it requires merge->getChild(0) to be a HASH_JOIN.

  If the identity “won’t be available until much later,” the robust version should not be “put dedup after the first hash join.” It should be:

  1. Find an UNWIND somewhere under MERGE.
  2. Find the shallowest operator under MERGE whose schema contains all merge.getKeys().
  3. Insert UNWIND_DEDUP above that operator.
  4. If no such operator exists before MERGE, do nothing.

  That generalizes to deep plans:

```
  MERGE
    A
      B
        C   <-- first point where all merge keys are in scope
          ...
            UNWIND
```

  rewritten as:

```
  MERGE
    A
      B
        UNWIND_DEDUP(keys = merge keys)
          C
            ...
              UNWIND
```

  The prototype here is a good proof that “dedup on real merge keys after they exist” works for the known bug. It is not the fully general insertion-point algorithm. For production, I’d prefer either the conservative fix already pushed, or invest in the generalized “first schema containing all merge keys”  rewrite with tests for nested plans.
  
  https://gist.github.com/adsharma/1b83118decf92270bf73a6b911ff201c

adsharma:
The conservative fix fails CI by hitting an assertion. So we have to go for the "more general" fix which reorders `UNWIND_DEDUP`. The updated PR does:

  - places UNWIND_DEDUP after the HASH_JOIN, closer to MERGE, so deeper matched identities can be in scope;
  - derives keys from merge.getKeys();
  - additionally includes relationship source/destination internal IDs for relationship MERGE;
  - disables that relationship rewrite when the MERGE is not the root operator.

Remaining work: making relationship MERGE dedup identity/state correct for deep or factorized plans, where it can be applied more broadly.

adsharma:

ericyuanhui:
> Thank you for the contribution!



> The conservative fix fails CI by hitting an assertion. So we have to go for the "more general" fix which reorders `UNWIND_DEDUP`. The updated PR does:
> 
> * places UNWIND_DEDUP after the HASH_JOIN, closer to MERGE, so deeper matched identities can be in scope;
> * derives keys from merge.getKeys();
> * additionally includes relationship source/destination internal IDs for relationship MERGE;
> * disables that relationship rewrite when the MERGE is not the root operator.
> 
> Remaining work: making relationship MERGE dedup identity/state correct for deep or factorized plans, where it can be applied more broadly.

I've seen commit 595fa48 merged into the master branch. Has all issues related to unwind merge commands been fully resolved here? @adsharma

adsharma:
@ericyuanhui  #170  and #378 are previous issues related to UNWIND. Looks like we forgot to add regression tests for those. They may have regressed. Needs testing.

Were you able to test your use case with the merged commit?

adsharma:
Confirmed that #170 was broken by this commit. Updated fix: #520

ericyuanhui:
> Confirmed that #170 was broken by this commit. Updated fix: #520

thank for your support. I will try the new fix to run my test case.

PR #512, opened 2026-05-23
Optimize LiveJournal degree count queries

adsharma:
## What Changed Summary

- Added `REL_DEGREE_TABLE` logical/physical source operator.
- q07 rewrite: `MATCH (u)-[:follows]->(v) RETURN count(DISTINCT u.id)` now becomes `REL_DEGREE_TABLE` in `ACTIVE_BOUND_COUNT` mode.
- q06 rewrite: `MATCH (u)-[:follows]->(v) RETURN u.id, count(v) AS deg ORDER BY deg DESC LIMIT 10` now becomes `REL_DEGREE_TABLE` in `TOP_K_DEGREES` mode after top-k optimization.
- Added CSR degree summary methods for native, icebug-disk/parquet CSR, and Arrow CSR rel tables.
- Kept q08 `COUNT_REL_TABLE` rewrite intact.

## Why

LiveJournal benchmark q06 and q07 were still executing scan-based count plans. These rewrites let the planner answer the unfiltered degree/count shapes from CSR metadata instead of scanning all `follows` edges.

## Validation

- `cmake --build build/release --target lbug_shell`
- LiveJournal icebug-disk benchmark smoke run with `build/release/tools/shell/lbug :memory: -b -i ../live-journal-benchmark/icebug_disk/schema.cypher`:
  - q06 returned the expected top 10 in ~48ms executing.
  - q07 returned `4004103` in ~0.94ms executing.
  - q08 returned `69362378` in ~0.09ms executing.
- Native in-memory smoke test for q06/q07 rewrite behavior.

### DIFF
--- a/src/processor/operator/scan/rel_degree_table.cpp
+++ b/src/processor/operator/scan/rel_degree_table.cpp
@@ -1,0 +1,105 @@
+#include "processor/operator/scan/rel_degree_table.h"
+
+#include <unordered_map>
+#include <unordered_set>
+
+#include "common/types/types.h"
+#include "main/client_context.h"
+#include "processor/execution_context.h"
+#include "transaction/transaction.h"
+
+using namespace lbug::common;
+using namespace lbug::storage;
+using namespace lbug::transaction;
+
+namespace lbug {
+namespace processor {
+
+void RelDegreeTable::initLocalStateInternal(ResultSet* resultSet, ExecutionContext*) {
+    degreeVector = resultSet->getValueVector(degreeOutputPos).get();
+    if (mode == planner::RelDegreeTableMode::TOP_K_DEGREES) {
+        nodeKeyVector = resultSet->getValueVector(nodeKeyOutputPos).get();
+    }
+    hasExecuted = false;
+}
+
+void RelDegreeTable::writeNodeKey(offset_t offset, sel_t pos) const {
+    switch (nodeKeyVector->dataType.getPhysicalType()) {
+    case PhysicalTypeID::INT32:
+        nodeKeyVector->setValue<int32_t>(pos, static_cast<int32_t>(offset));
+        break;
+    case PhysicalTypeID::INT64:
+        nodeKeyVector->setValue<int64_t>(pos, static_cast<int64_t>(offset));
+        break;
+    case PhysicalTypeID::UINT32:
+        nodeKeyVector->setValue<uint32_t>(pos, static_cast<uint32_t>(offset));
+        break;
+    case PhysicalTypeID::UINT64:
+        nodeKeyVector->setValue<uint64_t>(pos, static_cast<uint64_t>(offset));
+        break;
+    default:
+        nodeKeyVector->setValue<int64_t>(pos, static_cast<int64_t>(offset));
+        break;
+    }
+}
+
+bool RelDegreeTable::getNextTuplesInternal(ExecutionContext* context) {
+    if (hasExecuted) {
+        return false;
+    }
+    auto transaction = Transaction::Get(*context->clientContext);
+    if (mode == planner::RelDegreeTableMode::ACTIVE_BOUND_COUNT) {
+        if (relTables.size() == 1) {
+            const auto count = relTables[0]->getNumActiveBoundNodes(transaction, direction);
+            degreeVector->state->getSelVectorUnsafe().setToUnfiltered(1);
+            degreeVector->setValue<int64_t>(0, static_cast<int64_t>(count));
+            hasExecuted = true;
+            return true;
+        }
+        std::unordered_set<offset_t> activeOffsets;
+        for (auto* relTable : relTables) {
+            for (auto& [offset, _] : relTable->getDegreeEntries(transaction, direction)) {
+                activeOffsets.insert(offset);
+            }
+        }
+        degreeVector->state->getSelVectorUnsafe().setToUnfiltered(1);
+        degreeVector->setValue<int64_t>(0, static_cast<int64_t>(activeOffsets.size()));
+        hasExecuted = true;
+        return true;
+    }
+
+    std::unordered_map<offset_t, row_idx_t> degrees;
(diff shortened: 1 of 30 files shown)


adsharma:
How does it work?

### `COUNT_REL_TABLE`

  `COUNT_REL_TABLE` is the fast path for plain relationship cardinality.

  It replaces plans like:

```text
  SCAN_REL_TABLE -> PROJECT f._ID -> COUNT(f._ID)
```

  with a direct count over the relationship table.

  Instead of scanning every edge row, it asks the rel table for its row count. For icebug-disk/parquet-backed rel tables  this comes from table/file metadata. For native CSR tables it sums CSR lengths, with handling for committed/uncommitted rows and deletions.

  It returns one row: the total relationship count.

  Example:

```
  MATCH (a:user)-[f:follows]->(b:user)
  RETURN count(f);
```

  ### REL_DEGREE_TABLE

  `REL_DEGREE_TABLE` is the fast path for degree-derived count queries.

  It has two modes.

  #### ACTIVE_BOUND_COUNT

  Used for q07:

```
  MATCH (u:user)-[:follows]->(v)
  RETURN count(DISTINCT u.id);
```

  This counts how many source CSR rows have at least one outgoing edge, instead of scanning all edges and deduplicating  u.id.

  For icebug-disk CSR, this is:

```
  count rows where indptr[i + 1] > indptr[i]
```

  #### TOP_K_DEGREES

  Used for q06:

```
  MATCH (u:user)-[:follows]->(v)
  RETURN u.id, count(v) AS deg
  ORDER BY deg DESC
  LIMIT 10;
```

  This computes each source node’s degree from CSR row lengths, keeps a top-k heap, and outputs (u.id, deg) directly.

  For LiveJournal’s icebug-disk layout, this reads the CSR indptr degree summary instead of materializing ~69M follows rows,  aggregating, sorting, and limiting.

  ### Summary

  - `COUNT_REL_TABLE`: “How many edges are in this rel table?”
  - `REL_DEGREE_TABLE`: “How many bound nodes have edges?” or “Which bound nodes have the largest degrees?”

src/processor/operator/scan/rel_degree_table.cpp
```diff
@@ -0,0 +1,84 @@
...
+    if (entries.size() > limit) {
+        entries.resize(limit);
+    }
+    auto& selVector = degreeVector->state->getSelVectorUnsafe();
+    selVector.setToUnfiltered(entries.size());
+    if (nodeKeyVector) {
+        nodeKeyVector->state->getSelVectorUnsafe().setToUnfiltered(entries.size());
+    }
+    for (sel_t i = 0; i < entries.size(); ++i) {
+        writeNodeKey(entries[i].first, i);
```
aheev, line 97: shouldn't we sum for a node appearing in multiple tables?

src/storage/table/rel_table.cpp
```diff
@@ -551,6 +552,134 @@ row_idx_t RelTable::getNumTotalRows(const Transaction* transaction) {
...
+        }
+        for (offset_t i = 0; i < StorageConfig::NODE_GROUP_SIZE; ++i) {
+            pushDegree(groupStartOffset + i, degrees[i]);
+        }
+    }
+    if (transaction->isWriteTransaction()) {
+        if (auto* localTable = transaction->getLocalStorage()->getLocalTable(tableID)) {
+            auto& localRelTable = localTable->cast<LocalRelTable>();
+            for (const auto& [nodeOffset, rowIndices] : localRelTable.getCSRIndex(direction)) {
+                pushDegree(nodeOffset, rowIndices.size());
```
aheev: what if a nodeOffset  appears in both local and persistent?

src/storage/table/rel_table.cpp
```diff
@@ -551,6 +552,134 @@ row_idx_t RelTable::getNumTotalRows(const Transaction* transaction) {
...
+                hasRels[i] = hasRels[i] || csrIndex->getNumRows(i) > 0;
+            }
+        }
+        result += std::count(hasRels.begin(), hasRels.end(), true);
+    }
+    if (transaction->isWriteTransaction()) {
+        if (auto* localTable = transaction->getLocalStorage()->getLocalTable(tableID)) {
+            auto& localRelTable = localTable->cast<LocalRelTable>();
+            for (const auto& [_, rowIndices] : localRelTable.getCSRIndex(direction)) {
+                result += !rowIndices.empty();
```
aheev: ditto

aheev (changes requested):
Would love to see some tests involving active writes, multiple rel tables, bwd rel count

src/storage/table/arrow_rel_table.cpp
```diff
@@ -601,5 +602,60 @@ row_idx_t ArrowRelTable::getTotalRowCount(
...
+            result++;
+        }
+    }
+    return result;
+}
+
+std::vector<std::pair<offset_t, row_idx_t>> ArrowRelTable::getTopKDegreeEntries(
+    const transaction::Transaction* transaction, RelDataDirection direction, idx_t k) const {
+    if (layout != ArrowRelTableLayout::CSR || direction == RelDataDirection::BWD || k == 0) {
+        return const_cast<ArrowRelTable*>(this)->RelTable::getTopKDegrees(transaction, direction,
```
aheev: this would result in crash right? same for IceDiskRel tables

adsharma:
- Degree rows are now merged by bound node across persistent, committed in-memory, and local write-transaction
    data.
  - REL_DEGREE_TABLE now sums duplicate node degrees across multiple rel tables, and active-source counts
    deduplicate nodes across tables.
  - Arrow/IceDisk unsupported degree directions now return no fast-path rows instead of falling through into native
    CSR code.
  - Added optimizer coverage for active writes, multiple rel tables, and backward rel count.
  - Allowed the degree top-k rewrite for count(*) in the one-hop degree pattern.

aheev (approved):

aheev:
voops tests are failing

PR #515, opened 2026-05-24
Auto-load httpfs for remote shell init

adsharma:
## Summary
- auto-install and load httpfs before processing remote shell init files
- skip the auto-load when init is local or HTTPFS is already loaded
- skip ice-disk parquet version footer reads for remote paths during schema initialization

## Why
Remote init paths such as s3:// and xet:// need HTTPFS registered before the shell can open them. After that, ice-disk schema startup was still slow because table construction opened each remote parquet file just to check icebug_disk_version metadata. Remote version checks now defer by skipping that eager metadata read; local parquet files keep the compatibility check.

## Validation
- cmake --build build/llvm-release --target lbug_shell
- build/llvm-release/tools/shell/lbug --version
- local -i tools/shell/test/files/start.cypher smoke test
- /usr/bin/time -p build/llvm-release/tools/shell/lbug -i xet://datasets/ladybugdb/small-kgs/main/kg_history/icebug-disk/schema.cypher
- git diff --check

### DIFF
--- a/src/include/storage/table/ice_disk_utils.h
+++ b/src/include/storage/table/ice_disk_utils.h
@@ -10,6 +10,7 @@
 
 #include "common/constants.h"
 #include "common/exception/runtime.h"
+#include "common/file_system/local_file_system.h"
 #include "common/file_system/virtual_file_system.h"
 #include "processor/operator/persistent/reader/parquet/parquet_reader.h"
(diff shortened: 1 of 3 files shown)


adsharma:
Speeds up 

```
```

By avoiding many network roundtrips.

PR #516, opened 2026-05-24
Batch IceDisk relationship source scans

adsharma:
## Summary
```
lbug> MATCH (a:person)-[b:proposed]->(c:concept) RETURN *; 
<takes 17 secs because of many network roundtrips>
```

- batch IceDisk relationship source scans when the bound node scan has projected properties
- return IceDisk node scan rows in vector batches instead of one row per scan call

## Why
Remote relationship scans were slow when the bound node had projected properties: the planner could not use source-mode batching, and IceDisk node scans emitted one row per call, so the relationship parquet was rescanned once per bound node. IceDisk relationship extends can now scan the source node properties inside the source-mode rel operator, and IceDisk node scans emit vector batches.

## Validation
- profile MATCH (a:person)-[b:proposed]->(c:concept) RETURN *; via xet://datasets/ladybugdb/small-kgs/main/kg_history/icebug-disk/schema.cypher
  - before: SCAN_REL_TABLE ~9.8s, wall ~17s
  - after: SCAN_REL_TABLE ~3.7s, wall ~10s

### DIFF
--- a/src/include/processor/operator/scan/scan_rel_table.h
+++ b/src/include/processor/operator/scan/scan_rel_table.h
@@ -2,6 +2,7 @@
 
 #include "binder/expression/rel_expression.h"
 #include "common/enums/extend_direction.h"
+#include "processor/operator/scan/scan_node_table.h"
 #include "processor/operator/scan/scan_table.h"
 #include "storage/predicate/column_predicate.h"
 #include "storage/table/rel_table.h"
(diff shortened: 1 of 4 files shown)


aheev:
Maybe we can also make use of `prefetchMode` in parquet_reader?

> // TODO(Ziyi): We currently only support reading from local file system, thus the prefetch
    // mode is disabled by default. Add this back when we support remote file system.

adsharma:
Tried the prefetch mode, but it didn't help this particular use case. But agree we should enable.

aheev (approved):

PR #519, opened 2026-05-25
Use Release for Windows minimal CI

adsharma:
## Summary
- run the Windows minimal CI build in Release instead of RelWithDebInfo
- run Windows minimal tests from the Release test build
- revert the prior Windows shared/static workaround commits

## Why
The Windows minimal RelWithDebInfo job embeds MSVC debug info in object files. When the bundled static lbug.lib is produced, it exceeds the 4GB COFF archive limit and fails with LNK1248. Keeping this job on Release avoids the oversized archive while preserving the single static-library packaging model used by language bindings.

## Validation
CI runs in comments


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/26413843506
https://github.com/LadybugDB/ladybug/actions/runs/26413843599

PR #522, opened 2026-05-26
Enable parquet prefetch

adsharma:
## Summary
- enable Parquet scan prefetch by default now remote VFS paths support positioned reads
- keep skipped columns out of prefetch accounting and registration
- remove the stale local-filesystem-only TODO

## Validation
```
lbug> MATCH (a:person)-[b:proposed]->(c:concept) RETURN *;
(16 tuples)
(3 columns)
Time: 621.70ms (compiling), 8356.26ms (executing)
```

### DIFF
--- a/src/include/processor/operator/persistent/reader/parquet/parquet_reader.h
+++ b/src/include/processor/operator/persistent/reader/parquet/parquet_reader.h
@@ -32,9 +32,7 @@ struct ParquetReaderScanState {
     ResizeableBuffer defineBuf;
     ResizeableBuffer repeatBuf;
 
-    // TODO(Ziyi): We currently only support reading from local file system, thus the prefetch
-    // mode is disabled by default. Add this back when we support remote file system.
-    bool prefetchMode = false;
+    bool prefetchMode = true;
     bool currentGroupPrefetched = false;
 };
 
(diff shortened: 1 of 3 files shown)


aheev:
I will take a look at it tmrw. Seems like It would need changes on the thrift side as well

adsharma:
Ah right. I made the changes on the thrift side as well, but it didn't help on the `xet:/.../kg_history` database. So switched to batching (which worked) and threw away the previous work.

Let me look into recreating it.

adsharma:
The code had:

`  std::min(existing_head->GetEnd(), new_read_head.GetEnd()) - new_start`

  That computes the intersection-ish end, not the union end. Example:

  existing: [1000, 2000)
  new:      [1800, 2600)

  wrong merged range: [1000, 2000)
  correct range:      [1000, 2600)

With the old code, the new range could be partially dropped from the read-ahead buffer. Correctness of query  results usually survived because a later read that missed the prefetch buffer fell back to readFromFile, but the  prefetch registration no longer meant “this full range is buffered.” For remote files, that causes extra range reads and defeats part of prefetch.

adsharma:
The thrift change didn't change query performance. Improving further requires either changes to query plan or more aggressive local caching. Probably post 0.17.0 release work.

aheev (approved):

PR #527, opened 2026-05-27
Handle larger OS page sizes in VM eviction

### LINKED ISSUE #526: Buffer manager crashes on kernels with CONFIG_PAGE_SIZE_16KB=y (e.g., Raspberry Pi 5 aarch64)
## Summary

LadybugDB 0.16.1 crashes at runtime on kernels built with `CONFIG_PAGE_SIZE_16KB=y` (e.g., the official Raspberry Pi OS kernel for RPi 4/5). The buffer manager calls `madvise(addr, 4096, MADV_DONTNEED)` with 4 KB alignment, but on 16 KB page kernels `madvise` returns `EINVAL` for any address or length not aligned to the kernel page size (16 KB). This causes a fatal `RuntimeError` after approximately 10,000–15,000 write operations.

## System Information

```
OS:     Raspberry Pi OS (Debian bookworm), aarch64
Kernel: 6.18.29+rpt-rpi-2712 (official RPi kernel, CONFIG_PAGE_SIZE_16KB=y)
Python: 3.13
Package: ladybug==0.16.1
         ladybug-0.16.1-cp313-cp313-manylinux_2_26_aarch64.manylinux_2_28_aarch64.whl
```

**Key kernel configuration:**
```
CONFIG_PAGE_SIZE_16KB=y   # confirmed via /boot/config-$(uname -r)
```

```
$ getconf PAGE_SIZE
16384
```

## Minimal Reproduction

```python
import ladybug as lb

db   = lb.Database("/tmp/test.lbug", buffer_pool_size=256*1024*1024)
conn = lb.Connection(db)
conn.execute("CREATE NODE TABLE T(id STRING, PRIMARY KEY(id))")

# Fails at approximately i=12,000–15,000 (varies with node data size)
for i in range(20000):
    conn.execute(f"MERGE (:T {{id: 'node_{i}_{'x'*100}'}})")
```

**Error:**
```
RuntimeError: Buffer manager exception:
  Releasing physical memory associated with a frame failed with error code -1: Invalid argument.
```

## Root Cause

The buffer manager allocates its pool as:
```
mmap(NULL, 1073741824, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_NORESERVE, -1, 0)
```

When evicting a 4 KB frame, it calls:
```
madvise(frame_addr, 4096, MADV_DONTNEED)  → EINVAL on 16KB-page kernels
```

On Linux with `CONFIG_PAGE_SIZE_16KB=y`, the kernel page size is 16384 bytes. The `madvise(2)` syscall requires both `addr` and `length` to be aligned to the kernel page size. With 4 KB database frames, only every 4th frame happens to be 16 KB-aligned, so the vast majority of eviction calls fail.

**Python-level verification** (run on affected system):
```python
import ctypes, mmap

libc = ctypes.CDLL("libc.so.6", use_errno=True)

# 4KB-misaligned: fails on 16KB-page kernel
m = mmap.mmap(-1, 4096)
buf = (ctypes.c_char * 4096).from_buffer(m)
addr = ctypes.addressof(buf)
ctypes.set_errno(0)
ret = libc.madvise(ctypes.c_void_p(addr), ctypes.c_size_t(4096), ctypes.c_int(4))  # MADV_DONTNEED=4
print(ret, ctypes.get_errno())  # → -1  22  (EINVAL)

# 16KB-aligned: succeeds
m2 = mmap.mmap(-1, 16384)
buf2 = (ctypes.c_char * 16384).from_buffer(m2)
addr2 = ctypes.addressof(buf2)
ctypes.set_errno(0)
ret2 = libc.madvise(ctypes.c_void_p(addr2), ctypes.c_size_t(16384), ctypes.c_int(4))
print(ret2, ctypes.get_errno())  # → 0  0  (Success)
```

## Pattern of Failure

`strace -e madvise` shows LadybugDB calling `madvise` with 4096-byte frames:
```
madvise(0x7ffedbc51000, 4096, MADV_DONTNEED) = -1 EINVAL (Invalid argument)
```

The failure occurs consistently after frame ~1025, which is the point at which the buffer pool fills and the first eviction is attempted.

## Workaround Attempted

An `LD_PRELOAD` shim converting 16 KB-misaligned `MADV_DONTNEED`/`MADV_FREE` calls to no-ops suppresses the initial exception, but LadybugDB 0.16.1 then hits a secondary error:
```
RuntimeError: Buffer manager exception: Unable to allocate memory! The buffer pool is full and no memory could be freed!
```

This means the eviction path in 0.16.1 requires that `madvise` actually succeed and release physical memory. The no-op approach does not work for the current version.

## Suggested Fix

In the buffer manager's frame-release code, query the actual kernel page size at startup and round the `madvise` call up to the kernel page granule:

```cpp
// Before:
madvise(frame_addr, KUZU_PAGE_SIZE, MADV_DONTNEED);

// After:
static const size_t kernel_page = sysconf(_SC_PAGE_SIZE);
if (KUZU_PAGE_SIZE >= kernel_page) {
    // frame size >= page size: call is well-aligned by construction
    madvise(frame_addr, KUZU_PAGE_SIZE, MADV_DONTNEED);
} else {
    // frame size < page size: round addr down, length up
    uintptr_t aligned_addr = (uintptr_t)frame_addr & ~(kernel_page - 1);
    size_t aligned_len = kernel_page;  // covers at least one frame
    madvise((void*)aligned_addr, aligned_len, MADV_DONTNEED);
}
```

Alternatively, if `KUZU_PAGE_SIZE < sysconf(_SC_PAGE_SIZE)`, skip the `madvise` call entirely and accept that physical memory is not immediately returned (lower memory pressure matters less than correctness). A startup warning could inform users of the configuration mismatch.

## Impact

All LadybugDB users on:
- Raspberry Pi 4/5 running the official Raspberry Pi OS kernel
- Any Linux system with `CONFIG_PAGE_SIZE_16KB=y` or `CONFIG_PAGE_SIZE_64KB=y`

Raspberry Pi 5 is increasingly popular for AI/ML edge deployments. The RPi5 official kernel uses 16 KB pages for performance reasons, following Apple Silicon and other modern aarch64 SoCs. This bug makes LadybugDB entirely unusable on that platform for any workload that exceeds the buffer pool capacity.

adsharma:
Fixes #526.

## Summary

This changes VM eviction to distinguish Ladybug logical frame size from the operating system discard granularity. A build with 4 KiB database pages can run on a kernel with a larger page size, such as 16 KiB, without calling `madvise` on a non-OS-page-aligned 4 KiB frame.

`VMRegion` now detects the runtime OS page size and tracks resident logical frames per discard granule. The buffer manager charges the granule size only when the first logical frame in that granule becomes resident, and frees it only when the last logical frame in the granule is evicted.

## Root cause

The old VM region layout addressed frames as `region + frameIdx * LBUG_PAGE_SIZE` and evicted each frame with `madvise(frame, LBUG_PAGE_SIZE, MADV_DONTNEED)`. On a 16 KiB kernel page system, only every fourth 4 KiB frame begins on an OS page boundary, so `madvise` can fail for the other frames.

Aligning the `madvise` range outward would be unsafe because the surrounding 16 KiB OS page can contain neighboring logical frames that are still resident or dirty. This change waits until all logical frames in the discard granule are evicted before releasing the underlying OS page.

### DIFF
--- a/src/include/storage/buffer_manager/buffer_manager.h
+++ b/src/include/storage/buffer_manager/buffer_manager.h
@@ -275,12 +275,14 @@ private:
 
     uint64_t freeUsedMemory(uint64_t size);
 
-    void releaseFrameForPage(FileHandle& fileHandle [[maybe_unused]],
+    uint64_t releaseFrameForPage(FileHandle& fileHandle [[maybe_unused]],
         common::page_idx_t pageIdx [[maybe_unused]]) {
 #if BM_MALLOC
         // Page is freed instead in PageState::resetToEvicted
+        return fileHandle.getPageSize();
 #else
-        vmRegions[fileHandle.getPageSizeClass()]->releaseFrame(fileHandle.getFrameIdx(pageIdx));
+        return vmRegions[fileHandle.getPageSizeClass()]->releaseFrame(
+            fileHandle.getFrameIdx(pageIdx));
 #endif
     }
 
--- a/src/include/storage/buffer_manager/vm_region.h
+++ b/src/include/storage/buffer_manager/vm_region.h
@@ -1,6 +1,9 @@
 #pragma once
 
+#include <atomic>
+#include <memory>
 #include <mutex>
+#include <vector>
 
 #include "common/constants.h"
 #include "common/types/types.h"
@@ -22,8 +25,13 @@ public:
 
     common::frame_group_idx_t addNewFrameGroup();
 
+    // Mark the frame as resident. Returns the number of bytes that should be charged to the
+    // buffer pool for this frame.
+    uint64_t claimFrame(common::frame_idx_t frameIdx);
+
     // Use `MADV_DONTNEED` to release physical memory associated with this frame.
-    void releaseFrame(common::frame_idx_t frameIdx) const;
+    // Returns the number of bytes that should be released from the buffer pool accounting.
+    uint64_t releaseFrame(common::frame_idx_t frameIdx);
 
     // Returns true if the memory address is within the reserved virtual memory region
     bool contains(const uint8_t* address) const {
@@ -37,13 +45,21 @@ private:
     inline uint64_t getMaxRegionSize() const {
         return maxNumFrameGroups * frameSize * common::StorageConstants::PAGE_GROUP_SIZE;
     }
+    inline uint64_t getFrameGroupSize() const {
+        return static_cast<uint64_t>(frameSize) * common::StorageConstants::PAGE_GROUP_SIZE;
+    }
+    uint64_t getDiscardGranuleIdxInFrameGroup(common::frame_idx_t frameIdx) const;
+    uint64_t getFrameGroupIdx(common::frame_idx_t frameIdx) const;
 
 private:
     std::mutex mtx;
     uint8_t* region;
     uint32_t frameSize;
+    uint64_t discardGranuleSize;
+    uint64_t numDiscardGranulesPerFrameGroup;
     uint64_t numFrameGroups;
     uint64_t maxNumFrameGroups;
+    std::vector<std::unique_ptr<std::atomic<uint16_t>[]>> residentFramesPerDiscardGranule;
 };
 
 } // namespace storage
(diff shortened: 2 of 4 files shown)


adsharma:
Too risky for the 0.17.0 release. Let's get this in after the release goes out.

PR #534, opened 2026-05-28
add arrow csr dst col name param

aheev:
Adds support for custom destination column names in Arrow CSR relationship tables.

 Previously, CSR-backed relationship tables required the indices Arrow table to contain a destination offset column named `to`. This change adds a `dst_col_name` parameter to the C API/internal Arrow CSR creation path so callers can use a different column name, while preserving `"to"` as  the default when no name is provided.


aheev:
python: https://github.com/LadybugDB/ladybug-python/pull/21
rust: https://github.com/LadybugDB/ladybug-rust/pull/16

aheev:
@adsharma could you PTAL?

adsharma:
`ladybug-nodejs` needs an update too. We can deal with the rest later.

I'm looking into the pre-existing `capi` test failure. You can ignore that segfault for now.

adsharma (approved):

aheev:
nodejs PR: https://github.com/LadybugDB/ladybug-nodejs/pull/19

PR #536, opened 2026-05-28
Use trusted publishing for WASM deploy

adsharma:
## Summary
- remove the npm auth token from the WASM deploy job
- grant OIDC permissions and upgrade npm before publishing
- publish WASM packages with provenance while keeping the date-based nightly tag
- stop token-only dist-tag updates when a nightly package version already exists

## Validation
- ruby -e "require 'yaml'; YAML.load_file('.github/workflows/build-and-deploy.yml'); puts 'ok'"


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/26587862639

PR #542, opened 2026-05-30
add checks to block main graph create/drop

### LINKED ISSUE #541: Bug: nodejs open type graph error
### Ladybug version

v0.17.0

linux

### What happened?

Open type graphs seem unusable due to error about missing table.

Also would be nice if it were possible to make the default graph an open type, instead of having to create a graph and switch to it.

lbug-test.js:
```js
import lbug from "@ladybugdb/core";

async function main() {
  let db = new lbug.Database('example.lbug');
  let conn = new lbug.Connection(db);
  await conn.init();

  try {
    await conn.query(`CREATE GRAPH main ANY;`);
  } catch (e) {
    console.warn("failed to create graph 'main', it may already exist. Error:", { error: e });
  }
  await conn.query(`USE GRAPH main;`);
  const res1 = await conn.query(`CREATE (test:Test { a: 1 }) return test`);
  console.log("Result:", res1);
}

main().catch((err) => {
  console.error("Error in main:", err);
  process.exit(1);
});

```

```sh
node --experimental-default-type=module lbug-test.js
```

output:
```
> node --experimental-default-type=module lbug-test.js
Error in main: [Error: Binder exception: Table Test does not exist.]
```

aheev:
Closes https://github.com/LadybugDB/ladybug/issues/541

### DIFF
--- a/src/include/main/database_manager.h
+++ b/src/include/main/database_manager.h
@@ -33,11 +33,9 @@ public:
     void loadGraphsFromCatalog(storage::MemoryManager* memoryManager,
         main::ClientContext* clientContext);
     void setDefaultGraph(const std::string& graphName);
-    void clearDefaultGraph();
     bool hasGraph(const std::string& graphName);
-    catalog::Catalog* getGraphCatalog(const std::string& graphName);
     catalog::Catalog* getDefaultGraphCatalog() const;
-    bool hasDefaultGraph() const { return defaultGraph != "" && defaultGraph != "main"; }
+    bool hasDefaultGraph() const { return defaultGraph != ""; }
     std::string getDefaultGraphName() const { return defaultGraph; }
     std::vector<catalog::Catalog*> getGraphs() const;
     storage::StorageManager* getDefaultGraphStorageManager() const;
--- a/src/include/main/database_manager.h
+++ b/src/include/main/database_manager.h
@@ -33,9 +33,11 @@ public:
     void loadGraphsFromCatalog(storage::MemoryManager* memoryManager,
         main::ClientContext* clientContext);
     void setDefaultGraph(const std::string& graphName);
+    void clearDefaultGraph();
     bool hasGraph(const std::string& graphName);
+    catalog::Catalog* getGraphCatalog(const std::string& graphName);
     catalog::Catalog* getDefaultGraphCatalog() const;
-    bool hasDefaultGraph() const { return defaultGraph != ""; }
+    bool hasDefaultGraph() const { return defaultGraph != "" && defaultGraph != "main"; }
     std::string getDefaultGraphName() const { return defaultGraph; }
     std::vector<catalog::Catalog*> getGraphs() const;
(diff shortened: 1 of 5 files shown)


aheev:
curious, why were these checks added to drop operator class?

```
    auto dbManager = main::DatabaseManager::Get(*context);
    auto memoryManager = storage::MemoryManager::Get(*context);

    if (!dbManager->hasGraph(dropInfo.name)) {
        auto message = std::format("Graph {} does not exist.", dropInfo.name);
        switch (dropInfo.conflictAction) {
        case ConflictAction::ON_CONFLICT_DO_NOTHING: {
            appendMessage(message, memoryManager);
            return;
        }
        case ConflictAction::ON_CONFLICT_THROW: {
            throw BinderException(message);
        }
        default:
            UNREACHABLE_CODE;
        }
    }

    if (dbManager->hasDefaultGraph() && StringUtils::getUpper(dbManager->getDefaultGraphName()) ==
                                            StringUtils::getUpper(dropInfo.name)) {
        dbManager->clearDefaultGraph();
    }

    dbManager->dropGraph(dropInfo.name, const_cast<main::ClientContext*>(context));
    appendMessage(std::format("Graph {} has been dropped.", dropInfo.name), memoryManager);
```

it just calls dbManager functions, so, why not move this to dbManager and just call dbManager->dropGraph here?

adsharma:
Only the defaultGraph checking could be moved to the dbManager. The rest of it should stay in the drop operator because it's user facing cypher behavior.

aheev:
> Only the defaultGraph checking could be moved to the dbManager. The rest of it should stay in the drop operator because it's user facing cypher behavior.

can you re-trigger the build?

adsharma:
Can you create a more minimal diff?

```
 void DatabaseManager::createGraph(const std::string& graphName,
      storage::MemoryManager* memoryManager, main::ClientContext* clientContext, bool isAnyGraph) {
      if (StringUtils::caseInsensitiveEquals(graphName, "main")) {
          throw RuntimeException{"MAIN is a reserved graph name."};
      }

      // existing duplicate checks and create logic...
  }

  void DatabaseManager::dropGraph(const std::string& graphName, main::ClientContext* clientContext) {
      if (StringUtils::caseInsensitiveEquals(graphName, "main")) {
          throw BinderException{"Cannot drop the main graph."};
      }

      // existing catalog/in-memory drop logic...
  }
  ```
  
  Add the new tests for:

  - CREATE GRAPH main
  - CREATE GRAPH MAIN
  - CREATE GRAPH Main
  - CREATE GRAPH main ANY
  - DROP GRAPH main
  - DROP GRAPH IF EXISTS main
  - USE GRAPH main still works

  I would avoid these user visible changes:
  
  - changing hasDefaultGraph() semantics
  - changing setDefaultGraph("main") from "main" to ""
  - removing clearDefaultGraph()
  - removing getGraphCatalog()
  - removing “first created/loaded graph becomes default” behavior

adsharma:
Looks great!

PR #545, opened 2026-05-30
ci: disable wasm httpfs static link

adsharma:
## Summary

Temporarily disables static linking of the `httpfs` extension in the wasm package build.

## Context

The wasm build started failing after `httpfs` was added to `EXTRA_CMAKE_FLAGS` because the extension needs OpenSSL support. Initial attempts to satisfy that dependency showed that the available prebuilt wasm OpenSSL package is not compatible with `cpp-httplib` socket BIO usage, so this rolls wasm back to the previously working extension set while we sort out the right OpenSSL/httpfs approach.

## Validation

- `git diff --check`


adsharma:
This is a revert of #540

PR #547, opened 2026-05-31
Implement WAL group commits

adsharma:
## Summary

- add WAL group commit sequencing so multiple committers can share one durable sync
- split transaction commit into durable WAL append followed by in-memory publish
- preserve publish order and single-writer behavior while allowing committers to wait during sync

## Benchmark

C++ benchmark source: `node_write_bench.cpp`. Optimal was 64 threads at 6k writes/sec.

Setup:
- on-disk temp database under `/private/tmp`
- `enableMultiWrites=true`
- `autoCheckpoint=false`
- `forceCheckpointOnClose=false`
- one autocommitted `CREATE (:person {id, name})` per write
- one `Connection` per writer thread
- count verified after each run

### 1,000 writes per thread

| Branch | Threads | Total writes | Writes/sec |
| --- | ---: | ---: | ---: |
| main | 1 | 1,000 | 249.068 |
| group commit | 1 | 1,000 | 244.118 |
| main | 2 | 2,000 | 256.391 |
| group commit | 2 | 2,000 | 248.242 |
| main | 4 | 4,000 | 254.625 |
| group commit | 4 | 4,000 | 495.887 |
| main | 8 | 8,000 | 258.332 |
| group commit | 8 | 8,000 | 951.349 |

### 2,000 writes per thread

| Branch | Threads | Total writes | Writes/sec | Speedup |
| --- | ---: | ---: | ---: | ---: |
| main | 4 | 8,000 | 265.451 | 1.00x |
| group commit | 4 | 8,000 | 493.874 | 1.86x |
| main | 8 | 16,000 | 283.718 | 1.00x |
| group commit | 8 | 16,000 | 1,016.86 | 3.58x |
| main | 16 | 32,000 | 290.376 | 1.00x |
| group commit | 16 | 32,000 | 2,115.63 | 7.29x |

Main remains fsync-bound at roughly 250-290 writes/sec as writer concurrency increases. With group commits, throughput scales to roughly 1,017 writes/sec at 8 writers and 2,116 writes/sec at 16 writers for this benchmark.

### DIFF
--- a/src/binder/bind_expression/bind_parameter_expression.cpp
+++ b/src/binder/bind_expression/bind_parameter_expression.cpp
@@ -15,7 +15,7 @@ std::shared_ptr<Expression> ExpressionBinder::bindParameterExpression(
     auto& parsedParameterExpression = parsedExpression.constCast<ParsedParameterExpression>();
     auto parameterName = parsedParameterExpression.getParameterName();
     if (knownParameters.contains(parameterName)) {
-        return make_shared<ParameterExpression>(parameterName, *knownParameters.at(parameterName));
+        return make_shared<ParameterExpression>(parameterName, knownParameters.at(parameterName));
     }
     // LCOV_EXCL_START
     throw BinderException(
--- a/src/binder/expression/parameter_expression.cpp
+++ b/src/binder/expression/parameter_expression.cpp
@@ -17,7 +17,7 @@ void ParameterExpression::cast(const LogicalType& type) {
         // LCOV_EXCL_STOP
     }
     dataType = type.copy();
-    value.setDataType(type);
+    value->setDataType(type);
 }
 
 } // namespace binder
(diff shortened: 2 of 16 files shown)


adsharma:
Cc: @loganpowell

adsharma:
- WAL sync failures now still expose the allocated commit sequence to TM cleanup.
  - Exception cleanup now clears active write transaction state and wakes waiters.
  - Direct Transaction::commit() / publishCommit() now guard invalid commit timestamps.
  - Replaced the begin-transaction sleep loop with a condition variable.
  - Fixed WAL::reset() ordering.

adsharma:
Postgres and SQLite handle fsync() failures differently. Postgres poisons the WAL and blocks any future commits. SQLite continues with an "uncertain durability" state.

We're taking the Postgres approach.

PR #555, opened 2026-06-01
Add C# API submodule

sergey-v9:
Adds `tools/csharp_api` as the `LadybugDB/ladybug-dotnet` submodule.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "tools/wasm"]
 	path = tools/wasm
 	url = https://github.com/ladybugdb/ladybug-wasm
+[submodule "tools/csharp_api"]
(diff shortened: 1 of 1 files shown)


adsharma:
Not all languages do this. The only benefit of doing this is the convenience of running the release pipeline once and getting packages released in one shot. With more platforms/languages the pipeline becomes longer and longer and more fragile.

I'm not opposed to this, but would like to see a significant user base before adding it to the release pipeline.

Also, the trusted publisher flow points to submodule's `release.yml`, not `build-and-deploy.yml` in the monorepo.

sergey-v9:
Thanks! Got it. Then I'll close this, and it `ladybug-dotnet` will stay in its own repository.

PR #559, opened 2026-06-02
Fix Windows ARM Python wheel build

adsharma:
## Summary
- Make the liblbug download helper honor an explicit target architecture through `LBUG_ARCH`.
- Pass the Python wheel matrix platform into the helper before downloading precompiled liblbug.
- Remove the earlier static-only Windows archive workaround from the precompiled-binary workflow.

## Root cause
The Windows ARM Python wheel job was downloading the x86_64 static archive:

`Installed liblbug-static-windows-x86_64.zip ... on arm64`

The download helper derived architecture from `uname -m`. Under the Windows bash environment used by the workflow, that can report `x86_64` even while the wheel target is ARM64. The ARM64 wheel then tried to link against the x86_64 `lbug.lib`, producing unresolved/native link failures.

This was not a missing Arrow download. The job had selected the wrong precompiled Ladybug archive before the wheel link step.

## Validation
- Not tested locally per request.
- CI/build-and-deploy will validate Python wheels.


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/26834823202/job/79126880228

PR #571, opened 2026-06-06
add zizmor action

aheev:


aheev:
@adsharma there are a bunch of issues reported by zizmore in CI logs. Those will be added to `Security` tab after push to main

aheev:
For token issues, take a look at https://docs.github.com/en/actions/concepts/security/openid-connect

adsharma:
I'm not convinced that this needs to run on every PR. But let's try it and decide as we go.

aheev:
> I'm not convinced that this needs to run on every PR. But let's try it and decide as we go.

it only runs on PRs which have workflow changes

PR #582, opened 2026-06-11
Support secondary ART indexes

adsharma:
## Summary
- allow ART indexes on non-primary node properties
- store duplicate offsets in secondary ART leaves and maintain them across insert, update, delete, checkpoint, and reload
- use secondary ART indexes for simple single-table equality predicates

## Validation
- ninja -C build/release test/runner/e2e_test
- E2E_TEST_FILES_DIRECTORY=test/test_files build/release/test/runner/e2e_test --gtest_filter="*ddl_empty*"
- manual EXPLAIN check confirmed PRIMARY_KEY_SCAN_NODE_TABLE with Index: ART for a secondary name lookup


src/storage/index/art_index.cpp
```diff
@@ -409,40 +636,221 @@ std::unique_ptr<ArtPrimaryKeyIndex> ArtPrimaryKeyIndex::createNewIndex(IndexInfo
...
+        node->offset = offset;
+        return;
+    }
+    if (node->offset.value() == offset) {
+        return;
+    }
+    if (!node->overflowOffsets) {
+        node->overflowOffsets = std::make_unique<std::vector<offset_t>>();
+    }
+    if (std::find(node->overflowOffsets->begin(), node->overflowOffsets->end(), offset) ==
```
aheev: is this check needed?

src/storage/index/art_index.cpp
aheev, line 1: can this file be trimmed?

src/storage/index/art_index.cpp
adsharma: Converted to a DASSERT.

 That means if the same (key, offset) is inserted twice through a bug or replay path, the secondary ART leaf will contain duplicate offsets, and lookupAll can return the same node twice. Deletes/updates may also only erase one copy at a time.

For the current intended paths, duplicate insertion should not happen:

  - initial index build scans each row once
  - normal insert commits once
  - update does eraseOffsetInternal(root, offset) before reinserting
  - delete erases by offset

src/storage/index/art_index.cpp
adsharma: [6ac816a](https://github.com/LadybugDB/ladybug/pull/582/commits/6ac816a2f1b887ec2477e98bcd02733404b077df) splits the file logically. Not sure if you meant some code can be removed.

PR #591, opened 2026-06-16
alpine: use openssl, not openssl3

adsharma:
Fixes build issues.


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/27591150424/job/81573025595

PR #605, opened 2026-06-21
fix: resolve the deadlock by using atomics

adsharma:
a classic lock‑order inversion between two TransactionManager mutexes, hit because TransactionContext::commit() (when it triggers an auto/force checkpoint) and TransactionContext::beginAutoTransaction() (→ beginTransactionInternal() → TransactionManager::beginTransaction()) acquire them in opposite orders.

 The two locks involved:

 - mtxForSerializingPublicFunctionCalls ("public‑function" lock)
 - mtxForStartingNewTransactions ("write gate" / new‑transaction lock)

commit() calls checkpoint()/tryCheckpoint() → checkpointNoLock() when shouldAutoCheckpoint or shouldForceCheckpoint is set. So the cycle is:

 - Thread A (a writer whose commit() just fired an auto‑checkpoint): holds mtxForStartingNewTransactions (the write gate, kept after the drain loop exits), then tries to acquire mtxForSerializingPublicFunctionCalls to read lastTimestamp (line 305).
 - Thread B (another concurrent writer entering beginAutoTransaction() → beginTransaction(WRITE)): holds mtxForSerializingPublicFunctionCalls, then tries to acquire mtxForStartingNewTransactions (line 35).

### DIFF
--- a/src/include/transaction/transaction_manager.h
+++ b/src/include/transaction/transaction_manager.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <atomic>
 #include <condition_variable>
 #include <memory>
 #include <mutex>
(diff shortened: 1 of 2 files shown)


adsharma:
Tested by importing the ladybug source repo into lscope as of this commit: https://github.com/adsharma/lscope/commit/fd6cc127d079655c80df0c92cc7b2a4f593d6d37

```
python3 main.py --index /path/to/ladybug --language cpp --workers 8
Ingested 2602 file(s), 33273 semantic node(s), and 87726 resolved call(s) into test.db using 8 analysis thread(s)
  cpp: 2602 file(s)
       65.54 real        83.17 user         6.06 sys
          2065891328  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
              142518  page reclaims
                  18  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                   0  signals received
               17507  voluntary context switches
              950821  involuntary context switches
       1662480220213  instructions retired
        353909380349  cycles elapsed
          1539655816  peak memory footprint
 ```

PR #607, opened 2026-06-22
COPY and IGNORE_ERRORS with dups improvements

adsharma:
Address review comments from #601 

@ericyuanhui  could you please review/test this?


ericyuanhui:
> Address review comments from #601
> 
> @ericyuanhui could you please review/test this?

some test failed 
Assertion strings are outdated  
For example, the error message no longer mentions SKIP_DUPLICATE_PK, but the test is still matching the old wording exactly.  

Test input syntax is outdated  
Some test statements are still using SKIP_DUPLICATE_PK=true, while the goal of the current PR is to consolidate the user-visible syntax to just IGNORE_ERRORS=true (DUPLICATE_PK_ONLY).

ericyuanhui:
image
as this case . When I use IGNORE_ERRORS=true, shouldn't it execute according to the original semantics? At the very least, it should give the user some information that there was an error, and how to check it. But it seems like that's not the case. I did a test (IGNORE_ERRORS=true (DUPLICATE_PK_ONLY)), and currently that feature works fine. My previous design intent for the feature was to avoid affecting the original functionality and capabilities, after all, this feature is just an optimization for some special insert scenarios.

ericyuanhui:
image
 I try it with call show_warning. It will show it. So we just need add some waring log for user right?

PR #608, opened 2026-06-22
Fix segfault on `label()` of list element

mdbenito:
This PR fixes this segfault

```sql
import database 'dataset/demo-db/parquet';
MATCH p = (n)-[:LivesIn]->(m:City) WITH nodes(p) AS ns RETURN label(ns[1]);
```

And another one on comparisons like `ns[1] <> ns[0]`.

Here is the original bug report https://github.com/kuzudb/kuzu/issues/6050, addressed in Ryu with https://github.com/predictable-labs/ryugraph/pull/2

Despite the claim in #184 , this was never ported to ladybug, hence this PR.

The problem is incorrect binding of non-pattern exprs in `label()` and in comparisons. I added some new test cases for the latter.

Another two fixes were missing from that PR. I'll prepare separate ones for them

PR #609, opened 2026-06-22
Fix segfault when using ALL() for rel comparison

mdbenito:
This is a followup to #608 , completing the porting of the fixes in https://github.com/predictable-labs/ryugraph/pull/2.

Instead of the custom data in the original PR I used the tinysnb dataset.

I also added another change from that PR, defensively avoiding an assert in debug builds in `NodeOrRelExpression::addPropertyExpression`

### DIFF
--- a/src/binder/bind_expression/bind_comparison_expression.cpp
+++ b/src/binder/bind_expression/bind_comparison_expression.cpp
@@ -27,7 +27,14 @@ std::shared_ptr<Expression> ExpressionBinder::bindComparisonExpression(
     return bindComparisonExpression(parsedExpression.getExpressionType(), children);
 }
 
-static bool isNodeOrRel(const Expression& expression) {
+// A node or rel *pattern* expression (e.g. from a MATCH) has an internal ID property and can
+// be rewritten to an internal-ID comparison. A plain variable holding a NODE/REL value (e.g. a
+// lambda variable bound from `relationships(p)`) is a VariableExpression with a NODE/REL logical
+// type but no internal ID, so it must go through the regular value comparison path.
+static bool isNodeOrRelPattern(const Expression& expression) {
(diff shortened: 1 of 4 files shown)


adsharma:
```
[2026-06-22 14:55:30.271] [info] DEBUG LOG: RecursivePathWithAllAndRelComparison
[2026-06-22 14:55:30.271] [info] QUERY: MATCH (a:person)-[t:knows]->(d:person) WHERE a.ID <> d.ID WITH a, d, t MATCH p = (a)-[:knows*1..3]->(d) WHERE length(p) >= 2 AND ALL (r IN relationships(p) WHERE r <> t) WITH DISTINCT t RETURN count(t);
/home/runner/work/ladybug/ladybug/test/test_runner/test_runner.cpp:200: Failure
Value of: queryResult->isSuccess()
  Actual: false
Expected: true
Unexpected error for query: Assertion failed in file "/home/runner/work/ladybug/ladybug/src/include/common/cast.h" on line 30: false
```

mdbenito:
Yes, sorry, I thought I had rebased on the correct branch. Part of the fix was on my previous PR, so this one could not pass the test. Now it should be ok.

adsharma:
Something like:

https://gist.github.com/adsharma/9c189d8bec7df072cd3cae3ba30acd6d 

has the same effect but less repeated code.

mdbenito:
Actually there is no repetition, it's github being confused by the rebasing. This PR adds two lines. Let me rebase again on main

adsharma:
Yes, the duplicate code already went into main #608, but could be optimized via the change above, since the fall through code passes tests even after eliminating the 

```
if (child->expressionType == ExpressionType::PATTERN)
```

check inside the loop.

mdbenito:
Ahh, I see what you mean. Just check if it's a pattern outside so we don't repeat what comes after the loop. Done!

PR #610, opened 2026-06-22
fix(storage): guard CSR getGroup against stale/OOB group indices (SIGSEGV)

rysweet:
## Problem

A long-running process doing **relationship deletes + checkpoints** crashes with
`SIGSEGV` during the consolidation/checkpoint phase. The faulting access is
`GroupCollection<ChunkedNodeGroup>::getGroup(groupIdx = 4294967295 / UINT32_MAX)`.

### Root cause

After relationship delete + checkpoint churn, `csrIndex->indices[nodeOffset]` can
retain a **stale row** (an `INVALID_ROW_IDX` sentinel). `getQuotientRemainder(INVALID_ROW_IDX)`
yields an **out-of-range chunk-group index**. `GroupCollection::getGroup` only had a
debug-only `DASSERT(groupIdx < groups.size())` (a no-op in release), so it performed an
**out-of-bounds `std::vector` access** and callers dereferenced the result → undefined
behavior / `SIGSEGV`.

The insertion-write loop in `checkpointColumnInRegion` already guards this case with
`if (row == INVALID_ROW_IDX) continue;`, but the two **checkpoint-collect loops**
(`collectInMemRegionChangesAndUpdateHeaderLength` and `populateCSRLengthInMemOnly`),
plus the scan / `update` / `delete_` paths, do **not** — they dereference the `getGroup`
result directly.

## Fix

- `GroupCollection::getGroup` / `getGroupNoLock` now return `nullptr` for an out-of-range
  index instead of performing an OOB access.
- Every `CSRNodeGroup` deref site handles `nullptr`:
  - scan (`scanCommittedInMemSequential` / `scanCommittedInMemRandom`): skip / empty result;
  - `update`: no-op (target row gone); `delete_`: returns `false`;
  - checkpoint insertion-write: skip;
  - the two checkpoint-collect `isDeleted` loops: treat a stale/null group as **deleted**
    (purges the stale row via `turnToNonSequential()` + `setInvalid(i)`), consistent with
    the engine's existing handling of deleted rows.

This is version-independent (reproduced on 0.15.x–0.17.1) and present on current `main`.

## Validation

Reproduced via a deterministic 120-iteration delete+checkpoint churn test (amplihack-memory
issue #100). Before: `SIGSEGV` mid-churn in `collectInMemRegionChangesAndUpdateHeaderLength`.
After (engine rebuilt from source with this patch): the workload completes cleanly and the
test passes.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/include/storage/table/group_collection.h
+++ b/src/include/storage/table/group_collection.h
@@ -49,11 +49,19 @@ public:
     T* getGroup(const common::UniqLock& lock, common::idx_t groupIdx) const {
         DASSERT(lock.isLocked());
         UNUSED(lock);
-        DASSERT(groupIdx < groups.size());
+        // A stale/invalid CSR row can map to an out-of-range group index (e.g.
+        // getQuotientRemainder(INVALID_ROW_IDX)). Returning nullptr instead of
+        // performing an out-of-bounds std::vector access lets callers skip the
+        // stale row rather than dereferencing into undefined behavior (SIGSEGV).
+        if (groupIdx >= groups.size()) {
+            return nullptr;
+        }
         return groups[groupIdx].get();
     }
     T* getGroupNoLock(common::idx_t groupIdx) const {
-        DASSERT(groupIdx < groups.size());
+        if (groupIdx >= groups.size()) {
+            return nullptr;
+        }
         return groups[groupIdx].get();
     }
     void replaceGroup(const common::UniqLock& lock, common::idx_t groupIdx,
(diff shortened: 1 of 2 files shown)


rysweet:
Withdrawing this PR — it was opened prematurely. Closing to validate the fix and CI in a private fork first; will re-open a clean, CI-green PR if appropriate. Apologies for the noise.

PR #613, opened 2026-06-23
add lance tables support

aheev:


src/include/common/enums/storage_format.h
```diff
@@ -6,7 +6,7 @@
 namespace lbug {
 namespace common {
 
-enum class StorageFormat : uint8_t { NONE, ICEBUG_DISK };
+enum class StorageFormat : uint8_t { NONE, ICEBUG_DISK, LANCE };
```
adsharma: Does the base need to know about `LANCE`? I'm concerned that `isLance` kind of negates the purpose of extensions.

aheev: I don’t think it’s really possible to distinguish Lance from other tables without a catalog or namespace otherwise

aheev: are you concerned about the `mix tables` check or the `enum`? The enum serves a similiar purpose as `CatalogEntryType::FOREIGN_TABLE_ENTRY`

adsharma: Let's rename it to something generic instead of `Lance`. There are at least 5 table formats that want to replace parquet.

aheev:
finally!! all the tests are passing. Will refactor the changes and open it up for review tmrw

aheev:
Occupied with some personal work. I will get back to it shortly

PR #615, opened 2026-06-24
Fix read-only open during checkpoint from reading inconsistent state

ericyuanhui:
Summary
This change fixes a bug where opening the database in read-only mode could race with an in-progress checkpoint and end up reading inconsistent on-disk state. In that situation, recovery could observe intermediate checkpoint artifacts and continue loading from partially updated metadata, which could lead to crashes such as segmentation faults.
To address that, the fix makes checkpoint progress explicit and blocks read-only recovery from entering that unsafe window. It adds coordination around checkpoint execution so read-only opens can detect that a checkpoint is underway and fail fast instead of trying to recover from an intermediate state. It also adds defensive validation when reading checkpoint metadata so invalid or out-of-range page information is rejected with a controlled error rather than being used blindly. In addition, it cleans up a related resource-handling issue in file locking so failed lock attempts do not leak OS handles.
In short, the change turns a potentially unsafe concurrent recovery path into a safe, well-defined failure mode, and adds extra validation to prevent corrupted or partial checkpoint state from causing low-level crashes.

What Changed
Added two checkpoint lock files:*.checkpoint.intent.lock
*.checkpoint.apply.lock

Added checkpoint lock acquisition/release in Checkpointeracquire at checkpoint start
release on cleanup and rollback

Added eager creation of checkpoint lock files for non-read-only persistent databases
Added read-only recovery protection in WALReplayerif a checkpoint is in progress, opening in read-only mode now throws a RuntimeException
read-only recovery no longer proceeds through shadow file / checkpoint WAL intermediate states

Added defensive validation when reading checkpoint headersvalidate catalog page range
validate metadata page range
validate dataFileNumPages against actual file size

Fixed file descriptor / handle leaks when file locking fails in LocalFileSystem
Updated WAL testsReadOnlyRecoveryWithShadowFile now expects failure in read-only mode
added ReadOnlyRecoveryWithCheckpointWAL

### DIFF
--- a/src/include/storage/checkpointer.h
+++ b/src/include/storage/checkpointer.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <memory>
 #include <unordered_map>
 #include <vector>
 
@@ -15,6 +16,7 @@ namespace catalog {
 class Catalog;
 }
 namespace common {
+struct FileInfo;
 class VirtualFileSystem;
 } // namespace common
 namespace testing {
@@ -72,6 +74,8 @@ private:
 
     static void readCheckpoint(main::ClientContext* context, catalog::Catalog* catalog,
         StorageManager* storageManager);
+    void acquireCheckpointLocks();
+    void releaseCheckpointLocks();
 
     PageRange serializeCatalog(const catalog::Catalog& catalog, StorageManager& storageManager);
     PageRange serializeCatalogSnapshot(const catalog::Catalog& catalog,
@@ -105,6 +109,8 @@ protected:
         tableEpochWatermarksByManager;
     std::unordered_map<catalog::Catalog*, uint64_t> catalogVersionAtCheckpointByCatalog;
     std::unordered_map<StorageManager*, uint64_t> pageManagerVersionAtCheckpointByManager;
+    std::unique_ptr<common::FileInfo> checkpointIntentLockFile;
+    std::unique_ptr<common::FileInfo> checkpointApplyLockFile;
 };
 
 } // namespace storage
--- a/src/common/file_system/local_file_system.cpp
+++ b/src/common/file_system/local_file_system.cpp
@@ -123,6 +123,7 @@ std::unique_ptr<FileInfo> LocalFileSystem::openFile(const std::string& path, Fil
         BOOL rc = LockFileEx(handle, dwFlags, 0 /*reserved*/, 1 /*numBytesLow*/, 0 /*numBytesHigh*/,
             &overlapped);
         if (!rc) {
+            CloseHandle(handle);
             auto error = GetLastError();
             throw IOException("Could not set lock on file : " + fullPath +
                               " (Error: " + std::to_string(error) + ")\n" +
@@ -146,6 +147,7 @@ std::unique_ptr<FileInfo> LocalFileSystem::openFile(const std::string& path, Fil
         int rc = fcntl(fd, F_SETLK, &fl);
         if (rc == -1) {
             int original_errno = errno;
+            close(fd);
             if (original_errno == EAGAIN || original_errno == EACCES) {
                 struct flock get_fl {};
                 memset(&get_fl, 0, sizeof get_fl);
--- a/src/include/storage/storage_utils.h
+++ b/src/include/storage/storage_utils.h
@@ -77,6 +77,12 @@ public:
     static std::string getShadowFilePath(const std::string& path) {
         return std::format("{}.{}", path, common::StorageConstants::SHADOWING_SUFFIX);
     }
+    static std::string getCheckpointIntentLockFilePath(const std::string& path) {
+        return std::format("{}.checkpoint.intent.lock", path);
+    }
+    static std::string getCheckpointApplyLockFilePath(const std::string& path) {
+        return std::format("{}.checkpoint.apply.lock", path);
+    }
     static std::string getTmpFilePath(const std::string& path) {
         return std::format("{}.{}", path, common::StorageConstants::TEMP_FILE_SUFFIX);
     }
--- a/src/storage/checkpointer.cpp
+++ b/src/storage/checkpointer.cpp
@@ -1,8 +1,15 @@
 #include "storage/checkpointer.h"
 
+#include <chrono>
+#include <string_view>
+#include <thread>
 #include <vector>
 
 #include "catalog/catalog.h"
+#include "common/constants.h"
+#include "common/exception/io.h"
+#include "common/exception/runtime.h"
+#include "common/file_system/file_info.h"
 #include "common/file_system/file_system.h"
 #include "common/file_system/virtual_file_system.h"
 #include "common/serializer/buffered_file.h"
@@ -17,9 +24,11 @@
 #include "storage/database_header.h"
 #include "storage/shadow_utils.h"
 #include "storage/storage_manager.h"
+#include "storage/storage_utils.h"
 #include "storage/storage_version_info.h"
 #include "storage/wal/local_wal.h"
 #include "transaction/transaction.h"
+#include <format>
 
 namespace lbug {
 namespace storage {
@@ -63,6 +72,51 @@ void logCheckpointAndApplyShadowPagesForStorage(main::ClientContext& clientConte
     shadowFile.clear(*bufferManager);
 }
 
+bool isLockContention(const common::IOException& exception) {
+    return std::string(exception.what()).find("Could not set lock") != std::string::npos;
+}
+
+std::unique_ptr<common::FileInfo> acquireCheckpointWriteLock(main::ClientContext& clientContext,
+    const std::string& lockPath) {
+    auto vfs = common::VirtualFileSystem::GetUnsafe(clientContext);
+    while (true) {
+        try {
+            return vfs->openFile(lockPath,
+                common::FileOpenFlags(common::FileFlags::READ_ONLY | common::FileFlags::WRITE |
+                                          common::FileFlags::CREATE_IF_NOT_EXISTS,
+                    common::FileLockType::WRITE_LOCK),
+                &clientContext);
+        } catch (const common::IOException& exception) {
+            if (!isLockContention(exception)) {
+                throw;
+            }
+            std::this_thread::sleep_for(
+                std::chrono::microseconds(common::THREAD_SLEEP_TIME_WHEN_WAITING_IN_MICROS));
+        }
+    }
+}
+
+bool isValidCheckpointPageRange(PageRange range, common::page_idx_t numPages) {
+    if (range.startPageIdx == common::INVALID_PAGE_IDX) {
+        return range.numPages == 0;
+    }
+    if (range.numPages == 0 || range.startPageIdx >= numPages) {
+        return false;
+    }
+    return range.numPages <= numPages - range.startPageIdx;
+}
+
+void validateCheckpointPageRange(PageRange range, common::page_idx_t numPages,
+    std::string_view name) {
+    if (isValidCheckpointPageRange(range, numPages)) {
+        return;
+    }
+    throw common::RuntimeException(std::format(
+        "Cannot read checkpoint: {} page range starts at {} and spans {} pages, outside the "
+        "database file with {} pages. The database may be checkpointing; please retry later.",
+        std::string{name}, range.startPageIdx, range.numPages, numPages));
+}
+
 } // namespace
 
 Checkpointer::Checkpointer(main::ClientContext& clientContext)
@@ -71,6 +125,22 @@ Checkpointer::Checkpointer(main::ClientContext& clientContext)
       mainStorageManager{clientContext.getDatabase()->getStorageManager()} {}
 
 Checkpointer::~Checkpointer() = default;
+
+void Checkpointer::acquireCheckpointLocks() {
+    if (isInMemory || checkpointIntentLockFile || checkpointApplyLockFile) {
+        return;
+    }
+    const auto databasePath = clientContext.getDatabasePath();
+    checkpointIntentLockFile = acquireCheckpointWriteLock(clientContext,
+        StorageUtils::getCheckpointIntentLockFilePath(databasePath));
+    checkpointApplyLockFile = acquireCheckpointWriteLock(clientContext,
+        StorageUtils::getCheckpointApplyLockFilePath(databasePath));
+}
+
+void Checkpointer::releaseCheckpointLocks() {
+    checkpointApplyLockFile.reset();
+    checkpointIntentLockFile.reset();
+}
 
 std::vector<Checkpointer::CheckpointTarget> Checkpointer::collectCheckpointTargets() const {
     std::vector<CheckpointTarget> result;
@@ -157,6 +227,7 @@ void Checkpointer::writeCheckpoint() {
         return;
     }
 
+    acquireCheckpointLocks();
     checkpointTargets = collectCheckpointTargets();
 
     for (const auto& target : checkpointTargets) {
@@ -200,6 +271,7 @@ void Checkpointer::beginCheckpoint(common::transaction_t snapshotTimestamp) {
         return;
     }
 
+    acquireCheckpointLocks();
     snapshotTS = snapshotTimestamp;
     checkpointTargets = collectCheckpointTargets();
 
@@ -301,6 +373,7 @@ void Checkpointer::postCheckpointCleanup(bool canResetPageManagerToCurrent) {
         }
         target.storageManager->getShadowFile().reset();
     }
+    releaseCheckpointLocks();
 }
 
 bool Checkpointer::checkpointStorage() {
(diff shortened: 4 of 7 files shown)


src/include/storage/checkpointer.h
```diff
@@ -105,6 +109,8 @@ class Checkpointer {
         tableEpochWatermarksByManager;
     std::unordered_map<catalog::Catalog*, uint64_t> catalogVersionAtCheckpointByCatalog;
     std::unordered_map<StorageManager*, uint64_t> pageManagerVersionAtCheckpointByManager;
+    std::unique_ptr<common::FileInfo> checkpointIntentLockFile;
```
adsharma, line 112: Could you describe the thought process here? Why not lock the main database file instead of creating new intent/apply files?

adsharma:
Note that DuckDB and SQLite take different approaches here:

https://claude.ai/share/9ce5d5a8-bf34-46cb-87e7-9f1636491c4f

src/include/storage/checkpointer.h
ericyuanhui: the core reason why we cannot rely solely on the primary database file lock is straightforward: this fix targets the consistency window during checkpoints, rather than simply controlling which processes can open the primary data file.
Let’s break down the problem point by point.
1) Primary database locks cannot represent the phase semantics of a checkpoint
A primary file lock can only indicate whether a file is occupied. However, this bug requires us to distinguish three distinct states:
Normal readable state
Dangerous intermediate state while a checkpoint is in progress
Stable state after the checkpoint completes
This change introduces a two-phase locking mechanism: intent lock and apply lock. It explicitly marks the "checkpoint-in-progress" state instead of conflating it with generic file occupancy status. The relevant code is defined in storage_utils.h, and the locking logic is implemented in checkpointer.cpp.
2) Sole reliance on the primary lock leads to excessive blocking with overly coarse-grained semantics
If read-only open operations contend for the primary file lock to avoid race conditions, we end up with two undesirable outcomes:
All read-only openings fail whenever a writer process exists, resulting in overly conservative behavior.
Read-only access is permitted to proceed, yet we lack a way to precisely block only during the risky checkpoint window.
The goal of this patch is to block only unsafe time windows while avoiding unnecessary interruptions in other scenarios. This is why we introduced dedicated checkpoint coordination locks and status checks, rather than lumping all logic into the primary file lock.
3) The issue involves multi-file consistency, which a single-file lock cannot enforce
As highlighted in your summary, read-only recovery may encounter intermediate artifacts such as shadow files and checkpoint WAL entries.
The inconsistency window spans the main data file, shadow files and checkpoint WAL files. Locking only the primary file cannot verify whether the entire set of files is in an atomically recoverable state.
The corresponding safeguards are implemented in WALReplayer:
First perform checkpoint lock detection for read-only mode (implemented in wal_replayer.cpp).
Immediately trigger fail-fast errors for intermediate states involving shadow files or frozen WALs (also in wal_replayer.cpp).
4) Primary file locks become ineffective after process crashes
OS file locks are bound to the process lifecycle. Locks are automatically released when a process crashes, yet partial intermediate files may remain on disk.
To address this, the patch adds checks based on the existence of state files plus validation for metadata page ranges. Any leftover partial state after a crash will result in a controlled failure instead of blind reads that trigger segmentation faults.
Defensive validation for page ranges is added in the Checkpointer read path, converting invalid memory access crashes into managed exceptions. The code resides in checkpointer.cpp.
5) The commit essentially converts race conditions into protocol-defined failure modes
The final paragraph of your summary already captures the essence:
The fix does not aim to eliminate failures entirely. Instead, it ensures the system fails explicitly whenever it hits the unsafe window, eliminating undefined behavior.
This objective requires four components:
Explicit checkpoint coordination with intent and apply locks
Pre-emptive rejection logic for recovery (fail-fast on read-only opens)
Defensive validation of checkpoint metadata
Resource cleanup on lock acquisition failures to prevent file descriptor and handle leaks (implemented in local_file_system.cpp)
Conclusion:
The primary database file lock is a necessary but insufficient condition. It only enforces mutual exclusion, yet cannot implement the phase protocol, multi-file consistency boundaries, or detection of leftover intermediate states after crashes that this change requires.

ericyuanhui:
> Note that DuckDB and SQLite take different approaches here:
> 
> https://claude.ai/share/9ce5d5a8-bf34-46cb-87e7-9f1636491c4f

ladybug is same like SQlite right?

adsharma:
Do you use a filesystem that supports snapshots? From a maintainability and complexity perspective, I prefer a solution like this:

https://dev.to/shiviyer/step-by-step-implementation-filesystem-snapshots-in-postgresql-39ci

The SQLite method works even if the reader doesn't have write permissions to the dir where the database resides. Also our internals are different from SQLite and the development methodology is not as tightly controlled. That requires us to  operate with simpler abstractions.

ericyuanhui:
> Do you use a filesystem that supports snapshots? From a maintainability and complexity perspective, I prefer a solution like this:
> 
> https://dev.to/shiviyer/step-by-step-implementation-filesystem-snapshots-in-postgresql-39ci
> 
> The SQLite method works even if the reader doesn't have write permissions to the dir where the database resides. Also our internals are different from SQLite and the development methodology is not as tightly controlled. That requires us to operate with simpler abstractions.

What I'm asking is: is there anything wrong with my approach to this modification and implementation?
Looking at Ladybug's design philosophy — since Ladybug supports both read-write mode and read-only mode, if I need to have one connection accessing the same database file in read-write mode while another connection accesses it in read-only mode, then we'd end up with the intermediate state issue I mentioned earlier. That's why we need to add a locking mechanism to safeguard consistency.

adsharma:
Let me come back to this after the 0.18.x release. Requires more careful consideration.

ericyuanhui:
> Let me come back to this after the 0.18.x release. Requires more careful consideration.

new version is release. If we can discuss this PR?

adsharma:
Still looking to get 0.18.1 out with some bug fixes before letting in big PRs. But this is a good time to discuss the main ideas behind the PR and why it works.

adsharma:
None of these are an objection to the PR. Obviously what it's doing is useful based on feedback from @m0nkeyfl0wer and possibly others.

* What belongs to an embedded database vs what should be outside.

I would argue that network protocol implementation or a distributed consensus should exist in another library. Someone could then combine all 3 libraries to make a fully functional database.

* Value vs complexity trade-off

Certainly it provides value. But it also adds complexity. Which one is greater?

* Is there a strong theoretical basis for what it's doing?

I think the answer is yes, more on it in my next comment below.

* Are there other ways of doing this?

Many filesystems implement zero-copy snapshots. So one can do:

```
// sqlite
BEGIN EXCLUSIVE TRANSACTION;
zfs snapshot pool_name/dataset@snap_name
ROLLBACK TRANSACTION
```

or

```
// rocksdb
db->DisableFileDeletions();
// fs snapshot
db->EnableFileDeletions();
```

These will keep the complexity in the filesystem and/or orchestration between the fs snapshot and DB blocking of further writes.

The complexity containment is the #1 concern I have around this PR.

adsharma:
1. SQLite — closest mechanism match

 SQLite uses POSIX advisory byte-range locks on the database file itself with a five-state lock state machine:

 ```
   UNLOCKED → SHARED → RESERVED → PENDING → EXCLUSIVE
 ```

 - SHARED — multiple readers hold it concurrently
 - RESERVED — one writer, signaling intent to write; readers can still hold SHARED
 - PENDING — writer asks all current readers to leave; blocks new readers
 - EXCLUSIVE — full write lock, no readers remain

 The PR's intent.lock / apply.lock pair is conceptually the same two-phase idea as RESERVED → PENDING → EXCLUSIVE, just expressed as two separate files instead of byte-range offsets inside one file. Both designs explicitly model "first announce intent, then escalate to the apply/mutate phase." SQLite's lock state machine is documented in [docs/lockvar.tex](https://www.sqlite.org/lockvar/doc/trunk/www/lockvar.wiki) and src/pager.c in the SQLite source.

 2. Intention locks (Gray, 1975)

 The "intention lock" terminology comes from Jim Gray's paper "Granularity of Locks and Degrees of Consistency in a Shared Database" (1975/1976), which defined the hierarchical lock modes that every textbook DBMS teaches:

 - IS — Intention Shared
 - IX — Intention Exclusive
 - S — Shared
 - X — Exclusive
 - SIX — Shared Intention Exclusive

 The whole point of an "intention" lock is exactly what this PR does: signal at a coarse level that you intend to take a
 finer-grained exclusive lock soon, so other coarse-level lockers can coordinate without having to descend into the
 fine-grained tree. The PR doesn't descend into a hierarchy, but the naming and intent is straight out of that paper.

 3. ARIES (IBM) — two-phase prepare/commit

 The ARIES recovery algorithm (used in IBM Db2, PostgreSQL, SQL Server, and many others) has an explicit "prepare" phase (write intent record to log, acquire locks, flush) and a separate "commit" phase (write commit record, release locks). The "intent lock" terminology shows up in the ARIES/2PL implementations of these systems. The PR's "intent" and "apply" naming is reminiscent of that two-phase prepare/apply dance.

 4. PostgreSQL exclusive backup

 pg_basebackup's exclusive mode uses an exclusive lock on the database plus a backup_label file. A reader doing PITR/recovery detects the backup_label and refuses to start up (you have to remove it or use the non-exclusive mode). This is the same defensive "fail closed if intermediate state is on disk" idea the PR implements, but at the file-label level rather than a separate lock file.

adsharma:
### Races the PR does not cover

1. **Long-lived read-only connection vs. later writer activity** (the biggest gap)
   - The reader's lock files are released the moment `WALReplayer::replay()` returns. From that point on, the reader runs queries with no synchronization against the writer.
   - The writer can then run a checkpoint that calls `reclaimTailPagesIfNeeded(currentHeader->dataFileNumPages)`, **shrinking the data file**. The reader's buffer manager may have cached pages that are now beyond EOF, or it may read pages the writer has just overwritten.
   - The PR provides no MVCC / snapshot isolation for read-only — it only protects the *open* moment, not the *lifetime*.

2. **Reader opening the DB while a writer is appending to the active (non-frozen) WAL** (no checkpoint involved)
   - `WALReplayer::replay` reads the active WAL via `openWALFile()` with `READ_ONLY`. If the writer appends a new record concurrently, the reader can read a torn or partial record. The PR's `throwIfReadOnlyCheckpointState` does **not** trigger because the WAL is the active one, not the frozen one.
   - There is no lock file dedicated to "the writer is currently appending".

3. **Same-process reader and writer** (e.g., background threads, or two `Database` instances in one process)
   - POSIX `fcntl` locks are **per-process**, not per-file-descriptor. The writer's `WRITE_LOCK` and the reader's `READ_LOCK` on the same file will not contend in the same process.
   - This is a real concern for any architecture that puts a read-only view and a writer in the same process (replication followers, sidecar readers, embedded use).

4. **Stale lock files after a writer crash** (false positive, but worth flagging)
   - If the writer process is killed (`SIGKILL`, OOM, power loss) between acquiring the locks and `postCheckpointCleanup`, the lock files are released by the OS (good) but the shadow file / frozen WAL may be left behind. The next read-only open will hit `throwIfReadOnlyCheckpointState` and throw — a false positive that requires user retry. Safe, but not ideal.
   - There is no recovery path that distinguishes "writer crashed mid-checkpoint" from "writer is currently checkpointing" other than waiting and retrying.

5. **The data file is opened with `O_PERSISTENT_FILE_READ_ONLY` and without `O_LOCKED_PERSISTENT_FILE`** in read-only mode (`storage_manager.cpp:61-66`).
   - The reader can co-exist on the same data file as a writer that has the file locked. This is fine for *file content* (read-only is non-destructive) but it means the only mutual-exclusion channel is the lock files, which (a) only protect the checkpoint window and (b) are advisory on POSIX.

6. **No protection if the lock files themselves are deleted or corrupted**
   - Nothing prevents a user (or a misbehaving tool) from `rm`‑ing `*.checkpoint.intent.lock` mid-checkpoint. After that, the protection is gone with no warning.

7. **External tools that bypass `WALReplayer`**
   - Any tool that opens the data file directly (an offline analyzer, a backup tool, etc.) gets no protection. The lock files are only consulted by the in-process `WALReplayer`.

8. **Race against the catalog/metadata pages being rewritten even outside checkpoint**
   - DDL or metadata operations that update the catalog page range (e.g., `addProperty`, `createIndex`) write new catalog pages via the page manager. A reader that has already finished replay and is querying through cached buffers could observe torn writes. The PR adds validation only in `readCheckpoint`, not on the read path.

M0nkeyFl0wer:
> Do you use a filesystem that supports snapshots?

Data point from the deployment that produced the #642 repro: no. Plain ext4 on
Ubuntu 24.04 — no ZFS, no btrfs, no LVM thin pools. I'd guess that's the modal
environment for an embedded database's users: laptops, small VMs, default
distro filesystems. A solution that assumes zero-copy filesystem snapshots is
a solution most embedded users can't reach without reformatting.

What I actually deployed is the userland equivalent you're describing: the
writer owns the live file and, after each checkpoint drain (still holding its
write lock), refreshes a read-only snapshot copy that readers open exclusively.
It works — zero SIGSEGVs since — but the costs are real and worth putting on
the record for the complexity discussion:

- **A full file copy per write cycle.** My graph is 1.2 GB, so every drain
  cycle pays a 1.2 GB copy on ext4. Zero-copy would fix this, but see above.
- **Read staleness** between refreshes.
- **A failure mode the engine can't see.** The snapshot path, the atomic
  rename, the "never fall back to the live file" discipline — all of that
  lives in my application code, enforced by convention. Every new reader
  code path is a chance to reintroduce the crash. The engine can't help,
  because the invariant lives outside it.

That last point is my honest answer to the value-vs-complexity question. The
current contract is "read_only open racing a checkpoint is undefined behavior,
mitigate in application code," and the mitigation is subtle enough that I got
it wrong twice before getting it right: first with writer-side locking (which
can't protect a reader the engine lets open a mid-checkpoint file), then with
a snapshot scheme whose readers could silently fall back to the live file.
The version that works had to make the fallback a hard error — enforcing, in
application code, an invariant only the engine can actually see. Even if the full reader/writer
coordination story stays out of scope (your races 1–8 are a fair list, and
gap #1 in particular means this PR shouldn't be sold as general reader/writer
safety), there's a smaller contract worth buying: **a failed open never takes
down the process.** That's what #615 delivers on the open path, and it's the
difference between "retry loop" and "supervisor restarts my API."

On the "does this belong in the engine or outside" framing: SQLite's answer to
this exact question is instructive precisely because it put the locking state
machine *inside* — and that's a large part of why nobody rolls their own
concurrent-access layer around SQLite. The library that made "embedded" and
"safe concurrent access" compatible is the one that absorbed the complexity.

The #642 repro harness validates any revision of this fix in about a minute
(3/3 reader SIGSEGVs on stock 0.18.0, 0/6 with the #615 build). Happy to keep
running it against whatever shape the fix ends up taking, including a
minimal-scope version that only hardens the open path.

### Empirical check of race #1 (long-lived reader vs. writer checkpoints)

Since race #1 was flagged as the biggest gap, I tested it on stock 0.18.0
(PyPI): 3 reader processes each open the DB `read_only` **once**, hold the
connection for the whole run, and query in a tight loop, while a writer
process churns the same file. Five variants over a 200k-row seed, ~80 s each,
including a delete-heavy writer (every original row deleted and re-inserted,
~3.7M rows churned in one variant) and readers running full data scans with a
16 MiB buffer pool (far smaller than the file, forcing disk re-reads on every
scan).

Result: **it did not reproduce** — across ~1,700 writer checkpoints and
~780k reader queries, zero crashes, zero errors, zero wrong results. Every
long-lived reader served a perfectly consistent open-time snapshot for its
entire lifetime and exited 0.

The named mechanism also appears unreachable in 0.18.0: the data file never
shrank in any run. Reading the source, `reclaimTailPagesIfNeeded`
(`src/storage/page_manager.cpp:71`) adds tail pages to the FreeSpaceManager
for reuse but nothing truncates the main data file (only the WAL is
truncated), so "reader's cached pages beyond EOF" can't happen on this
version. Caveat: this is a null result, not a proof — the overwritten-page
variant of the hazard could still exist under free-page reuse patterns my
writers didn't generate (the file grew rather than recycling the reader's
exact pages).

Two side observations from the same runs, possibly worth their own issues:
(1) checkpoint never returns disk space in 0.18.0 — delete-heavy workloads
only grow the file; (2) a held read-only connection is frozen at its
open-time snapshot forever — it never observes any writer commit, so
"long-lived reader" also means "arbitrarily stale reader." If that's the
intended semantics it would be worth documenting, because it means the
lifetime-coordination problem in race #1 is currently more about staleness
than memory safety.

If useful, I can file the race-#1 lifetime question as a separate tracking
issue so this PR's scope can stay cleanly "make the open race fail closed."

adsharma:
> A solution that assumes zero-copy filesystem snapshots is
a solution most embedded users can't reach without reformatting.

Agreed. I found other arguments you make in the comment also compelling.

This fixes a real problem. As long as we don't sell it as a complete solution for reader-writer synchronization, we're good to go.

Yes, please open an issue for tracking race #1.

adsharma:
I'll merge this after the 0.18.1 release is made.

PR #624, opened 2026-06-27
optimization: skip merging arrow results if ordering is not specified

adsharma:
This helps avoid copies and sorting when scanning a rel table without ORDER BY.


adsharma:
This approach doesn't work. Redesigning and opening a new PR.

PR #629, opened 2026-06-28
fix: python build wheels on macos-15

adsharma:
Should fix this: https://github.com/LadybugDB/ladybug/actions/runs/28321832817/job/83908424071


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/28331533495

PR #634, opened 2026-07-01
fix: make java and CI

adsharma:


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/28489317966

PR #640, opened 2026-07-02
Fix SSL CA certificate detection for extension downloads

mdbenito:
Extension downloads over HTTPS fail with SSL verification errors when a Ladybug wheel built on one Linux distro (e.g. RHEL) is run on another (e.g. Debian). The vendored httplib (v0.14.2) falls back to OpenSSL's compile-time default CA paths, which don't match the runtime system's CA bundle location.

This PR adds `ExtensionUtils::getCaCertPath()` which probes for a usable CA bundle in order: `SSL_CERT_FILE`/`SSL_CERT_DIR` env vars, then well-known distro paths (Debian, RHEL, openSUSE, Alpine, macOS Homebrew), then CApath directories. The result is wired into the extension download client via `set_ca_cert_path()`. If nothing is found, it returns `std::nullopt` and OpenSSL's defaults apply (preserving existing behavior).

## Limitation

macOS Keychain loading is not supported by the vendored httplib v0.14.2 (that support was added in later upstream versions). On a clean macOS install without a Homebrew-provided CA bundle at `/etc/ssl/cert.pem`, the probe will find nothing and fall back to OpenSSL defaults, which may still fail. Probably the real fix for macOS is upgrading the vendored httplib.

### DIFF
--- a/src/extension/extension.cpp
+++ b/src/extension/extension.cpp
@@ -1,6 +1,8 @@
 #include "extension/extension.h"
 
+#include <array>
 #include <cstdlib>
+#include <filesystem>
 
 #include "common/exception/io.h"
 #include "common/string_utils.h"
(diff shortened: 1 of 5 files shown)


mdbenito:
gh runner ran out of space https://github.com/LadybugDB/ladybug/actions/runs/28604605424/job/84821529263?pr=640#step:13:1808

PR #649, opened 2026-07-04
Fix O(N²) ALP metadata recomputation in ColumnChunkData::split()

### LINKED ISSUE #645: Importing FLOAT/DOUBLE extremely slow (450x) due to alp::AlpEncode<> being slow
### Ladybug version

v0.18.0

ubuntu 26.04 lts, kernel 7.0

### What happened?

While benchmarking Ladybug on an LDBC SNB-derived graph (2.9M nodes, 6.1M directed
edges across 3 node tables and 3 rel tables), I measured a large asymmetry between
node and relationship bulk loading with `COPY ... FROM` CSV:

| load | rows | time | rate |
|---|---|---|---|
| all 6 node tables | 2.9M | 3.2 s | ~900K rows/s |
| KNOWS rel table | 346K | 205 s | ~1.7K rows/s |
| CREATED rel table | 2.86M | 1468 s | ~2.0K rows/s |
| HAS_TAG rel table | 2.93M | 1615 s | ~1.8K rows/s |

The rel-COPY rate is invariant to everything I could vary, which suggests the cost is
in the relationship-copy execution path itself rather than I/O or my usage pattern:

- `parallel=true` vs `parallel=false`: marginal difference (e.g. 3.1 s vs 6.9 s on a
  node table; rel tables ~unchanged).
- One `COPY` with a file list vs sequential COPYs of 250K-row chunks: 187 s vs 205 s
  on KNOWS (so it is not repeated adjacency merging across COPY statements).
- **On-disk vs `:memory:` database: 187–205 s vs 197 s on KNOWS — identical**, so it
  is not buffer-manager or disk I/O bound.

Schema shape: rel tables have 13 property columns (STRING/INT64/DOUBLE/BOOLEAN);
node tables have 15. Node primary keys are INT64. The graph has a skewed degree
distribution (social-network-like; max out-degree ~500 at this scale).

Repro sketch (same data pipeline as #475):

```python
import ladybug
db = ladybug.Database("db_dir", max_db_size=1 << 43)   # also repros with :memory:
conn = ladybug.Connection(db)
# CREATE NODE TABLE / REL TABLE per schema, then:
conn.execute('COPY Person FROM "person.csv" (header=true, parallel=true)')   # fast
conn.execute('COPY KNOWS  FROM "knows.csv"  (header=true, parallel=true)')   # ~2K rows/s
```

At this rate a 78M-edge graph takes ~11 hours to load, which currently keeps Ladybug
out of the larger scale points of the benchmark in which it otherwise participates.

Happy to provide the exact CSVs/generator (same reproducibility package as #475) or
run diagnostics against a branch.

_No response_

adsharma:
Fixes: #645 

```
# kuzu 0.11.3
--- 7. Create REL table (Owns) with 500,000 edges via COPY ---
  COPY completed in 0.25s (0.49 µs/edge)
```

vs

```
# https://github.com/kuzudb/kuzu/commit/9f5acab2e  + this commit
--- 7. Create REL table (Owns) with 500,000 edges via COPY ---
  COPY completed in 2.19s (4.37 µs/edge)
```

So it's slower due to segmentation and better estimation of how the DOUBLE column would compress, but not by 450x.

The Segmentation commit (895091340) introduced split(), which called getSizeOnDiskInMemoryStats() every 64 values in a tight inner loop. For DOUBLE/FLOAT columns, each call triggers the full ALP compression analysis (find_top_k_combinations + createFloatMetadata), creating O(N²) behavior: ~512 metadata recomputations per 256KB segment instead of one.

Fix this by computing the compression metadata once for the whole chunk, then deriving a count-based bound (maxPages × valuesPerPage) to use as the inner loop condition for fixed-size types. The bound respects page boundaries and avoids the integer-division bias of a simple avg-bytes-per-value approach. A single verification call per segment confirms the segment fits within targetSize.

Variable-size types (strings, lists) don't trigger expensive ALP analysis, so they keep the original metadata-based check.

### DIFF
--- a/src/storage/table/column_chunk_data.cpp
+++ b/src/storage/table/column_chunk_data.cpp
@@ -1057,8 +1057,32 @@ std::vector<std::unique_ptr<ColumnChunkData>> ColumnChunkData::split(bool target
     // the size above MAX_SEGMENT_SIZE, since this will still sometimes produce segments larger than
     // MAX_SEGMENT_SIZE
     auto maxSegmentSize = std::max(getMinimumSizeOnDisk(), common::StorageConfig::MAX_SEGMENT_SIZE);
-    auto targetSize =
-        targetMaxSize ? maxSegmentSize : std::min(getSizeOnDisk() / 2, maxSegmentSize);
+
+    // Compute compression metadata once. This runs expensive analysis (e.g., ALP's
+    // find_top_k_combinations for float types) exactly once for the whole chunk, instead of O(N)
+    // times in the inner loop below.
+    auto meta = getMetadataToFlush();
+    auto dataPages = meta.getNumDataPages(dataType.getPhysicalType());
+    uint64_t nullSize = 0;
+    if (nullData) {
+        nullSize = nullData->getSizeOnDisk();
+    }
+    auto totalSize = dataPages * common::LBUG_PAGE_SIZE + nullSize;
+
+    auto targetSize = targetMaxSize ? maxSegmentSize : std::min(totalSize / 2, maxSegmentSize);
+
+    // For fixed-size types, compute a count-based bound directly from the compression metadata.
+    // Using pages * valuesPerPage is accurate because it respects page boundaries and avoids the
+    // integer-division bias of a simple avg-bytes-per-value approach.
+    uint64_t maxValuesPerSegment = UINT64_MAX;
+    if (numBytesPerValue > 0 && numValues > 0 && dataPages > 0) {
+        const auto valuesPerPage = common::ceilDiv(numValues, static_cast<uint64_t>(dataPages));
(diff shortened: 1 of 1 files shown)


adsharma:
Why take a 10x hit on `COPY`? Here's the likely motivation for introducing segmentation:

Data is now stored in independently-manageable segments rather than one monolithic chunk. This enables:

 1. More granular memory management — segments can be individually spilled to disk, flushed, or checkpointed.
 2. Better I/O — checkpointing can write only the affected segments instead of the entire column.
 3. Finer control over data size — MAX_SEGMENT_SIZE prevents any single segment from growing unbounded, which is important for databases handling large tables with wide columns.

PR #663, opened 2026-07-07
Fix wrong CASE WHEN values for bool references

### LINKED ISSUE #662: Bug: CASE WHEN with bare boolean reference returns incorrect results on multi-row input
### Ladybug version

latest

macos and linux

### What happened?

`CASE WHEN` returns incorrect results when the `WHEN` condition is a bare boolean reference on multi-row
  input.

  The issue does not appear when the input contains only one row, and it also does not appear when the
  boolean reference is compared explicitly with `true`.

  This suggests the problem is likely in the vectorized selection path for boolean reference expressions.

  ## Reproduction

  ```cypher
  UNWIND [false, true] AS metric_0_value
  WITH *, CASE WHEN metric_0_value THEN "yes" ELSE "no" END AS label_value
  RETURN metric_0_value, label_value;
  ```
  Actual result
  ```
  ┌────────────────┬─────────────┐
  │ metric_0_value │ label_value │
  │ BOOL           │ STRING      │
  ├────────────────┼─────────────┤
  │ False          │ yes         │
  │ True           │ no          │
  └────────────────┴─────────────┘
  ```
  Expected result
  ```
  ┌────────────────┬─────────────┐
  │ metric_0_value │ label_value │
  │ BOOL           │ STRING      │
  ├────────────────┼─────────────┤
  │ False          │ no          │
  │ True           │ yes         │
  └────────────────┴─────────────┘
  ```
  Single-row sanity check

  The single-row case works correctly:

  ```
  UNWIND [true] AS metric_0_value
  WITH *, CASE WHEN metric_0_value THEN "yes" ELSE "no" END AS label_value
  RETURN metric_0_value, label_value;
  ```
  Actual result:
  ```
  True | yes
  ```
  Workaround

  Using an explicit comparison works correctly:

  ```
  UNWIND [false, true] AS metric_0_value
  WITH *, CASE WHEN metric_0_value = true THEN "yes" ELSE "no" END AS label_value
  RETURN metric_0_value, label_value;
  ```
  Expected/actual result:
  ```
  False | no
  True  | yes
  ```
 ##  Possible root cause

  The likely root cause is in the selection-vector handling for bare boolean reference expressions.

  A suspected area is:
  ```
  src/expression_evaluator/reference_evaluator.cpp
  ```
  In` ReferenceExpressionEvaluator::selectInternal()`, selected positions appear to be written into the
  result vector's own selection vector:

`  auto selectedBuffer = resultVector->state->getSelVectorUnsafe().getMutableBuffer();`

  instead of the SelectionVector& selVector passed by the caller.

  This can corrupt or desynchronize the caller-owned selection vector used by CASE WHEN.

  In particular, `CaseExpressionEvaluator::evaluate() `calls:

```
  auto whenSelVector = alternativeEvaluator.whenSelVector.get();
  auto hasAtLeastOneValue = alternativeEvaluator.whenEvaluator->select(*whenSelVector, false);

```
  and then later uses that same whenSelVector to fill the THEN branch values:

`  fillSelected(*whenSelVector, thenVector);`

  If ReferenceExpressionEvaluator::selectInternal() updates resultVector->state->getSelVectorUnsafe()
  instead of the passed-in whenSelVector, the CASE evaluator may fill the THEN result into stale or
  incorrect row positions.

  That would explain why this query:

```
  UNWIND [false, true] AS metric_0_value
  WITH *, CASE WHEN metric_0_value THEN "yes" ELSE "no" END AS label_value
  RETURN metric_0_value, label_value;
```

  produces:

```
  False | yes
  True  | no

  instead of:

  False | no
  True  | yes
```

  A possible fix may be to write selected positions into the passed-in selVector:

```
  auto selectedBuffer = selVector.getMutableBuffer();
  and to treat resultVector->state->getSelVector() only as the input selection vector.
```

_No response_

Mrhs121:
close https://github.com/LadybugDB/ladybug/issues/662

### DIFF
--- a/src/expression_evaluator/reference_evaluator.cpp
+++ b/src/expression_evaluator/reference_evaluator.cpp
@@ -13,15 +13,16 @@ inline static bool isTrue(ValueVector& vector, uint64_t pos) {
 
 bool ReferenceExpressionEvaluator::selectInternal(SelectionVector& selVector) {
     uint64_t numSelectedValues = 0;
-    auto selectedBuffer = resultVector->state->getSelVectorUnsafe().getMutableBuffer();
-    if (resultVector->state->getSelVector().isUnfiltered()) {
-        for (auto i = 0u; i < resultVector->state->getSelVector().getSelSize(); i++) {
+    auto selectedBuffer = selVector.getMutableBuffer();
+    auto& resultSelVector = resultVector->state->getSelVector();
+    if (resultSelVector.isUnfiltered()) {
+        for (auto i = 0u; i < resultSelVector.getSelSize(); i++) {
             selectedBuffer[numSelectedValues] = i;
             numSelectedValues += isTrue(*resultVector, i);
         }
     } else {
-        for (auto i = 0u; i < resultVector->state->getSelVector().getSelSize(); i++) {
-            auto pos = resultVector->state->getSelVector()[i];
+        for (auto i = 0u; i < resultSelVector.getSelSize(); i++) {
+            auto pos = resultSelVector[i];
             selectedBuffer[numSelectedValues] = pos;
             numSelectedValues += isTrue(*resultVector, pos);
         }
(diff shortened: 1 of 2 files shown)


adsharma:
Thank you!

PR #676, opened 2026-07-09
Remove global npm upgrade from deploy-nodejs/deploy-wasm jobs

adsharma:
The 'npm install -g npm@latest' step causes MODULE_NOT_FOUND errors for the 'sigstore' module needed by --provenance during npm publish. This is a known issue where npm 11's in-place global upgrade on GitHub hosted runners fails to properly lay down its restructured dependency tree (sigstore is missing from node_modules).

The bundled npm 10.9.x in both Node 20 and Node 22 already supports --provenance (added in npm 9.5.0), so the upgrade is unnecessary.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -419,9 +419,6 @@ jobs:
           node-version: "22"
           registry-url: "https://registry.npmjs.org"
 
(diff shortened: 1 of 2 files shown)


adsharma:
https://github.com/LadybugDB/ladybug/actions/runs/29061053787/job/86262792934

PR #689, opened 2026-07-15
fix: stop creating checkpoint lock files in StorageManager constructor

adsharma:
The checkpoint intent/apply lock files (*.checkpoint.intent.lock and *.checkpoint.apply.lock) were being created as empty placeholders by every StorageManager constructor, including for subgraph databases (e.g. test.mygraph1.db).  These files are unnecessary because all access (including subgraph access) goes through the main database path via clientContext.getDatabasePath().

The files are already created on demand by Checkpointer:: acquireCheckpointLocks() (with CREATE_IF_NOT_EXISTS) when a checkpoint actually starts, and cleaned up by releaseCheckpointLocks() after the checkpoint completes.  Pre-creating them in the StorageManager constructor was redundant and caused stale files to leak on disk for subgraph databases that are checkpointed via collectCheckpointTargets() but never have their own locks acquired or released.


adsharma:
Cc: @ericyuanhui

ericyuanhui:
Looks perfect! No issues with this modification at all. Thank you so much.

PR #704, opened 2026-07-19
fix(extension): name the missing dependency when Windows load fails

### LINKED ISSUE #685: Bug: [Windows] FTS extension 0.18.1 requires OpenSSL DLLs that are not distributed
### Ladybug version

0.18.1

Windows 11 x64

## Problem

On a clean Windows installation, the FTS extension cannot load because it depends on two OpenSSL 3 libraries that are not distributed with either `@ladybugdb/core` or the downloaded extension:

- `libssl-3-x64.dll`
- `libcrypto-3-x64.dll`

`INSTALL fts` successfully downloads `libfts.lbug_extension`, but `LOAD EXTENSION fts` fails with Windows error 126: “The specified module could not be found.”

Inspection of the extension's import table confirms that it imports both OpenSSL DLLs.

## Diagnostic workaround

Prepending an existing OpenSSL 3 installation to `PATH` makes the same extension load successfully. For example, Git for Windows happens to contain compatible DLLs on some machines:

```powershell
$env:Path = "C:\Program Files\Git\mingw64\bin;$env:Path"
node .\repro.mjs
```

This confirms the missing dependency, but relying on another application's private DLLs is not a safe deployment solution.

## Expected behavior

The Windows FTS extension should work on a clean supported Windows installation.

Possible fixes include:

1. Distribute the required OpenSSL DLLs beside `libfts.lbug_extension`.
2. Statically link OpenSSL into the Windows extension.
3. If users must install OpenSSL themselves, document the requirement and report the exact missing DLL rather than the generic Windows error 126.

A clean-Windows CI test should install and load FTS without inheriting OpenSSL from Git, Conda, or another tool's `PATH`.

## Impact

Applications using LadybugDB cannot reliably enable FTS on Windows. Behavior currently depends on unrelated software having placed compatible OpenSSL DLLs on the machine's `PATH`.

## Related issues

- LadybugDB Windows FTS loading crash: https://github.com/LadybugDB/ladybug/issues/431
- Downstream diagnosis confirming the missing DLLs: https://github.com/abhigyanpatwari/GitNexus/issues/2374
- Similar OpenSSL packaging problem on macOS: https://github.com/LadybugDB/ladybug/issues/681

```js
// repro.mjs
import lbug from "@ladybugdb/core";

const db = new lbug.Database("./fts-repro.lbug");
const conn = new lbug.Connection(db);

try {
  await conn.query("INSTALL fts");
  await conn.query("LOAD EXTENSION fts");
  console.log("FTS loaded successfully");
} catch (error) {
  console.error(error);
  process.exitCode = 1;
} finally {
  await conn.close();
  await db.close();
}
```

```powershell
npm install @ladybugdb/core@0.18.1

where.exe libssl-3-x64.dll
where.exe libcrypto-3-x64.dll

node .\repro.mjs
```

On a clean machine, both `where.exe` checks report that the files cannot be found and the extension fails to load.

freya0926:
Windows reports a missing transitive dependency as `ERROR_MOD_NOT_FOUND` against the library passed to `LoadLibraryW`, so `LOAD EXTENSION fts` on a machine without OpenSSL 3 fails with "the specified module could not be found" — naming the extension, which is present. Users are left guessing which module is actually meant (#685).

When the extension file itself exists, this adds a line saying the missing module is something the extension links against, and names the OpenSSL 3 DLLs.

**Scope:** this only improves the message. It does not fix #685 — @adsharma confirmed the core links OpenSSL dynamically, which means the core has already loaded `libssl-3-x64.dll` before any extension loads, so the fix for the underlying packaging problem belongs elsewhere.

**Testing:** compiles clean on MSVC 14.44 (VS 2022 Build Tools, Windows SDK 10.0.26100), full static Release build links. I could not exercise the message at runtime — `lbug_shell.exe` built here crashes with `0xC0000409` on database open, which I confirmed is pre-existing by reproducing it with this change reverted.

### DIFF
--- a/src/extension/extension.cpp
+++ b/src/extension/extension.cpp
@@ -369,13 +369,38 @@ void ExtensionUtils::registerIndexType(main::Database& database, storage::IndexT
     database.getStorageManager()->registerIndexType(std::move(type));
 }
 
+#ifdef _WIN32
+// Windows reports a dependency it cannot find as ERROR_MOD_NOT_FOUND against the library we asked
+// for, so the message names the extension while the file actually missing is something the
+// extension links against. Say so, rather than leaving users to guess which module is meant.
+static std::string dlMissingDependencyHint(const std::string& path) {
+    // Capture the code before touching the filesystem, which would overwrite it -- the caller still
+    // needs GetLastError() intact for dlErrMessage().
+    const auto errorCode = GetLastError();
+    const auto extensionExists = std::filesystem::exists(path);
+    SetLastError(errorCode);
+    if (errorCode != ERROR_MOD_NOT_FOUND || !extensionExists) {
+        return "";
+    }
+    return "\nThe extension itself was found, so a library it depends on is the one that could not "
+           "be located. Extensions link against OpenSSL 3 (libssl-3-x64.dll and "
+           "libcrypto-3-x64.dll), which is not distributed with them.";
+}
+#else
+static std::string dlMissingDependencyHint(const std::string&) {
+    return "";
+}
+#endif
+
 ExtensionLibLoader::ExtensionLibLoader(const std::string& extensionName, const std::string& path)
     : extensionName{extensionName} {
     libHdl = dlopen(path.c_str(), RTLD_NOW | RTLD_LOCAL);
     if (libHdl == nullptr) {
+        // Read before dlErrMessage(), which is also driven by GetLastError().
+        const auto hint = dlMissingDependencyHint(path);
         throw common::IOException(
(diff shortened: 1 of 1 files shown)


adsharma:
Closing based on no bots policy. Humans generating AI assisted PRs is ok.

freya0926:
Thanks for the clarification — I removed the "Generated with Claude Code" footer from the description above. To be clear: I'm a human, and I used Claude Code as a tool to help draft this fix, the same way one might use any other coding assistant. Happy to answer questions about the change itself if useful.

PR #709, opened 2026-07-21
Fix labels()/label() returning _nodes/_edges on ANY graphs

### LINKED ISSUE #708: Bug: labels()/label() return internal table name "_nodes" on ANY graphs instead of the node's labels
### Ladybug version

`0.19.0` (built from source, commit `70468ff1f`). Also reproduces on the `0.18.2` PyPI wheel.

### Operating system

macOS 15 (Darwin 25.3.0), arm64

### What happened?

In an `ANY` graph, `labels(n)` (and its alias `label(n)`) return the internal storage table name `_nodes` instead of the node's actual label set.

The stored data is correct — `RETURN n.*` returns the proper label list — so only `labels()`/`label()` read the wrong source.

- **Expected:** `labels(n)` → `[Person, Employee]` (matching `n.*` and Neo4j semantics)
- **Actual:** `labels(n)` → `_nodes`

### Steps to reproduce

```cypher
CREATE GRAPH g ANY;
USE GRAPH g;
CREATE (:Person:Employee {name: 'Alice'});
MATCH (n) RETURN n.*;        // [Person,Employee]   ✅ correct
MATCH (n) RETURN labels(n);  // _nodes              ❌ wrong
```

### Root cause (traced) & offer to fix

In `LabelFunction::rewriteFunc` (`src/function/pattern/label_function.cpp`), an ANY-graph node has a single catalog entry — the internal `_nodes` table — so `isMultiLabeled()` is `false` and the function returns `getEntry(0)->getName()`, i.e. `"_nodes"`. The real labels live in the per-node `label STRING[]` column of `_nodes` (the same column `n.*` reads).

I'd be happy to open a PR. The detector `isAnyGraphNodeOrRel(...)` already exists in `bind_property_expression.cpp`, and an ANY-graph node already carries the `label` property, so the fix can route `labels()` to that column for ANY nodes. One wrinkle worth noting: `rewriteFunc`'s result also feeds the node's `_LABEL` struct field (typed `STRING`), so the query path and the struct-field path need to be disambiguated to keep `RETURN n` type-safe.

### Two questions before I do

1. `labels()` (plural) and `label()` (singular) are aliased through the same function. On ANY graphs both would return a `STRING[]` list, whereas structured `label()` returns a scalar. Should the singular form stay scalar?
2. Rels have the identical bug (`label(e)` → `_edges`). Would you prefer a node-only fix scoped to this report, or node + rel together?

zachwinter:
## Description

Fixes #708. On an ANY ("open type") graph, `labels(n)`/`label(n)` returned the internal `_nodes` table name (and `label(e)`/`labels(e)` returned `_edges`) instead of the node/rel's actual labels. The data was already correct — `RETURN n.*` returns the real list — only these functions read the wrong source: an ANY node/rel has a single catalog entry (the `_nodes`/`_edges` table), so `isMultiLabeled()` is false and `LabelFunction::rewriteFunc` returned the table name.

Per your guidance on the issue:
- `labels(x)` → the full `STRING[]` list; `label(x)` → `labels(x)[0]` (first label, scalar), not throw.
- Fixed for **both node and rel**.

### Approach
`label()` and `labels()` alias through the same `LabelFunction::rewriteFunc`, and the resolved function name is `"LABEL"` for both — only the catalog-lookup name (`bindRewriteFunctionExpression`, already computed) distinguishes them. I thread that name into `RewriteFunctionBindInput` as a defaulted field so the shared rewrite can branch on singular vs plural. On an ANY graph it returns the per-row `label` column (the same one `n.*` reads) instead of the table name:

- node `label STRING[]`: `labels(n)` → the list; `label(n)` → `list_extract(label, 1)`.
- rel `label STRING` (scalar): `label(e)` → the scalar; `labels(e)` → wrapped into a 1-element list.

The **empty-name default** keeps the two precompute sites (`createQueryNode` / `createNonRecursiveQueryRel`, which materialize the `_LABEL` struct field) on their existing path, so `RETURN n` / `RETURN e` are unchanged, and structured graphs are fully unaffected — the new branch is gated on a non-empty name **plus** `isAnyGraphNodeOrRel(...)`.

### Tests
Added `AnyGraphLabelsFunctions` to `test/test_files/graph/any.test` covering node + rel, singular + plural. Existing `function/label.test` (structured `tinysnb`) and the rest of `graph/any.test` still pass locally (`e2e_test`, RelWithDebInfo).

### Known follow-up (out of scope, flagging so it reads as deliberate)
`RETURN n` / `RETURN e` whole-object materialization still shows `_nodes`/`_edges` in the `_LABEL` position — that struct field is typed `STRING`, and changing it touches scan/materialization. Happy to address separately if you'd like.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] I have changed storage version if on disk format has changed. — N/A, no on-disk format change
- [ ] I have updated the documentation (if needed). — ANY graphs appear undocumented today; happy to add docs in a separate PR.

### DIFF
--- a/src/binder/bind_expression/bind_function_expression.cpp
+++ b/src/binder/bind_expression/bind_function_expression.cpp
@@ -142,7 +142,7 @@ std::shared_ptr<Expression> ExpressionBinder::bindRewriteFunctionExpression(
         entry->ptrCast<FunctionCatalogEntry>());
     auto function = match->constPtrCast<RewriteFunction>();
     DASSERT(function->rewriteFunc != nullptr);
-    auto input = RewriteFunctionBindInput(context, this, children);
+    auto input = RewriteFunctionBindInput(context, this, children, functionName);
     return function->rewriteFunc(input);
 }
 
--- a/src/binder/bind_expression/bind_property_expression.cpp
+++ b/src/binder/bind_expression/bind_property_expression.cpp
@@ -34,8 +34,7 @@ static bool isStructPattern(const Expression& expression) {
            logicalTypeID == LogicalTypeID::STRUCT;
 }
 
-static bool isAnyGraphNodeOrRel(const NodeOrRelExpression& nodeOrRel,
-    main::ClientContext* context) {
+bool isAnyGraphNodeOrRel(const NodeOrRelExpression& nodeOrRel, main::ClientContext* context) {
     auto transaction = transaction::Transaction::Get(*context);
     auto useInternal = context->useInternalCatalogEntry();
     auto dbManager = main::DatabaseManager::Get(*context);
--- a/src/function/pattern/label_function.cpp
+++ b/src/function/pattern/label_function.cpp
@@ -7,6 +7,7 @@
 #include "catalog/catalog_entry/table_catalog_entry.h"
 #include "function/binary_function_executor.h"
 #include "function/list/functions/list_extract_function.h"
+#include "function/list/vector_list_functions.h"
 #include "function/rewrite_function.h"
 #include "function/scalar_function.h"
 #include "function/schema/vector_node_rel_functions.h"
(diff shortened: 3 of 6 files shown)


adsharma:
Looks good. Thank you for the contribution!

Issue #712, opened 2026-07-21
ANY graph: RETURN n / RETURN e expose _LABEL as internal table name (_nodes/_edges)


zachwinter:
Follow-up to #708 / #709 (now merged — thanks!). That PR fixed the `labels()`/`label()` *functions*; this is the whole-object counterpart I flagged there, which is a small design question more than a bug, so raising it before sending a PR.

### Ladybug version
Current `main` (`cf875edfc`); behavior is independent of the #709 fix.

### What happens
On an ANY graph, whole-object `RETURN n` / `RETURN e` expose the system `_LABEL` field as the **internal table name** (`_nodes` / `_edges`) rather than the node/rel's real label(s). The correct labels *are* present — as the `label` data column — so only `_LABEL` is stale:

```cypher
CREATE GRAPH g ANY;
USE GRAPH g;
CREATE (:Person:Employee {name:'Alice'})-[:WorksAt {since:2020}]->(:Org {name:'Acme'});

MATCH (n:Person:Employee) RETURN n;
// {_ID:…, _LABEL: '_nodes',  id:0, label:['Person','Employee'], data:'{"name":"Alice"}'}
//                  ^^^^^^^ internal table name           ^^^^^^^^^^^^^^^^^^^ correct labels

MATCH ()-[e:WorksAt]->() RETURN e;
// {_SRC:…, _DST:…, _LABEL: '_edges', _ID:…, label:'WorksAt', data:'{"since":2020}'}
```

It propagates into paths (`RETURN p`) too — every node's `_LABEL` is `_nodes`.

### The design question
`_LABEL` is a scalar `STRING` (`getBaseNodeStructFields`, `bind_graph_pattern.cpp:252`), but an ANY node can carry **multiple** labels (`_nodes.label` is `STRING[]`). So making `RETURN n` correct forces a choice about what `_LABEL` should be:

| Option | `RETURN n` `_LABEL` | Notes |
|---|---|---|
| **A. `_LABEL` → `STRING[]` for ANY nodes** | `['Person','Employee']` | Most faithful; consistent with `labels()`. Changes the NODE logical type for ANY graphs, so consumers of `_LABEL` (C API, Arrow, path/result formatting) must accept a list. |
| **B. `_LABEL` → first label (scalar)** | `'Person'` | Type-safe and minimal, but lossy (silently drops `Employee`). |
[...]

PR #719, opened 2026-07-22
Hide ANY graph internal columns from whole-object output

zachwinter:
## Description

Part of #712, and a follow-on to #715. On an ANY graph, whole-object `RETURN n` / `RETURN e` leaked the internal `_nodes`/`_edges` columns as if they were user properties — even though the struct already exposes the same information as `_ID` / `_LABEL`. This removes that redundancy as a deliberate first step toward the North Star (output identical to a strongly-typed subgraph).

Before → after (ANY graph only):
```
RETURN n :  {_ID, _LABEL, id, label, data}          ->  {_ID, _LABEL, data}
RETURN e :  {_SRC,_DST,_LABEL, _id, label, data}    ->  {_SRC,_DST,_LABEL, _id, data}
```

`id` duplicates `_ID`; `label` duplicates `_LABEL` (surfaced in #715). `data` — the JSON property bag — is kept; promoting it to top-level typed fields is the larger follow-on, which I've written up as a design question on #712 (it needs a property registry + a type policy that's really a maintainer call).

### Approach
The internal columns must stay **bound** — the insert path (`bindInsertNode`/`bindInsertRel`) looks them up by name, and explicit `n.id` access resolves through them — so they aren't dropped. Instead `NodeOrRelExpression` gains a `hiddenPropertyNames` set and a `getProjectedPropertyExpressions()` accessor (full list minus hidden). The whole-object struct type is built from that projected list, and `ExpressionMapper::getNodeEvaluator` / `getRelEvaluator` consume the same list so the struct fields stay positionally aligned with the child expressions that fill them.

In `createQueryNode` / `createNonRecursiveQueryRel` the ANY and structured paths are separate branches of a single if/else, with the shared struct assembly factored into `getNodeStructFields(node, labelType)` / `getRelStructFields(rel)`:
- **ANY**: mark the internal columns hidden (`{id, label}` for nodes, `{label}` for rels), build the struct, point `_LABEL` at the real `label` column.
- **structured**: build the struct, resolve `_LABEL` via the rewrite.

Structured graphs — which may legitimately have user columns named `id`/`label`/`data` — are completely untouched (empty hidden set ⇒ projected list == full list).

### Scope
Deliberately narrow — this only changes whole-object output:
- **`.*` is unchanged.** Hiding `label` there would remove the only per-row type signal, since star expansion has no `_LABEL` to fall back on and an ANY result can mix labels. `a.*` still returns `id | label | data`.
- **`show_tables()` is unchanged.** The other half of "hide `_nodes`/`_edges` in public APIs" is a genuine API-design choice (name-filter vs `isInternal=true` catalog entries, plus consistent behaviour when a graph's backing file is opened directly), which I raised on #712. Happy to follow up once the direction is set.

### Tests
Updated the two whole-object cases in `test/test_files/graph/any.test`. Structured-graph tests (`typed_subgraph.test`, `label.test`) and the path / recursive-join suites are unaffected — proving the ANY gate.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x] I have changed storage version if on disk format has changed. — N/A, no on-disk format change
- [ ] I have updated the documentation (if needed). — ANY graphs still undocumented; separate docs PR.

### DIFF
--- a/test/test_files/graph/any.test
+++ b/test/test_files/graph/any.test
@@ -25,8 +25,8 @@
 -LOG QueryAllNodes
 -STATEMENT MATCH (a) RETURN a.*
 ---- 2
-0|[User]|{"age":30,"name":"Alice"}
-1|[City]|{"population":8000000,"name":"New York"}
+{"age":30,"name":"Alice"}
+{"population":8000000,"name":"New York"}
 
 -LOG QueryDynamicProperties
 -STATEMENT MATCH (u:User)-[:LivesIn]->(c:City) RETURN u.name, c.name
@@ -46,7 +46,7 @@ Alice|30
 -LOG QueryDynamicNodePropertyInWhereFilter
 -STATEMENT MATCH (u:User) WHERE u.age = 30 RETURN u.*
 ---- 1
-0|[User]|{"age":30,"name":"Alice"}
+{"age":30,"name":"Alice"}
 
 -LOG QueryDynamicStringNodePropertyInWhereFilter
 -STATEMENT MATCH (u:User) WHERE u.name = 'Alice' RETURN u.name, u.age
@@ -98,7 +98,7 @@ Alice|30
 -LOG MatchConjunctiveLabelsReturnNodeStar
 -STATEMENT MATCH (n:Person:Employee) RETURN n.*
 ---- 1
-0|[Person,Employee]|{"name":"Alice"}
+{"name":"Alice"}
 
 -LOG Cleanup
 -STATEMENT USE GRAPH main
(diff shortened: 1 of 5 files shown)


test/test_files/graph/any.test
```diff
@@ -25,8 +25,8 @@
 -LOG QueryAllNodes
 -STATEMENT MATCH (a) RETURN a.*
 ---- 2
-0|[User]|{"age":30,"name":"Alice"}
-1|[City]|{"population":8000000,"name":"New York"}
+{"age":30,"name":"Alice"}
```
adsharma: Not sure if this is desirable. It's important for the user to figure out the type of the row.

adsharma (commented):
Does it make sense to separate the `isAny` paths from the `!isAny` path so they're not so intertwined?

test/test_files/graph/any.test
zachwinter: > Not sure if this is desirable. It's important for the user to figure out the type of the row.

Ah! Apologies, I assumed whole-object and .* were symmetric projections of the same property list.

zachwinter:
> Does it make sense to separate the `isAny` paths from the `!isAny` path so they're not so intertwined?

Aye, done in `4f89c865a`.
 
`isAny` is gone as a variable. Each mode is now a self-contained branch of a single `if`/`else` in both `createQueryNode` and `createNonRecursiveQueryRel`:
 
* ANY: mark the internal columns hidden, build the struct, point `_LABEL` at the real label column.
* structured: build the struct, resolve `_LABEL` via the rewrite.
 
The shared struct assembly moved into `getNodeStructFields(node, labelType)` / `getRelStructFields(rel)` so neither branch open-codes it. Previously the flag leaked into four places (the hidden-names call, a `_LABEL` type ternary, and the label branch) — now it's one predicate at the top.

PR #722, opened 2026-07-22
Row-Driven Primary-Key Lookup for MATCH

ericyuanhui:
## Problem

Queries that match a node by a primary key derived from an upstream row cannot use the existing
`PRIMARY_KEY_SCAN_NODE_TABLE` optimization because the key is not a constant expression. For
example:

```cypher
UNWIND $rows AS row
MATCH (e:Event {id: row.event_id})
MATCH (a:Auth {id: row.auth_id})
MERGE (e)-[:REL]->(a);
```

Previously, each dynamic `MATCH` was planned as an independent node-table scan plus filter and
combined with the upstream input. This can scan the full target node table and process a large
cross product even when each input row identifies at most one node by primary key.

The existing child-driven `IndexLookup` operator cannot be reused directly: it is designed for
relationship `COPY FROM`, reports missing endpoints as COPY errors, and builds a full in-memory
PK-to-offset cache when the target table does not have a persistent PK index.

## Change

This PR introduces a dedicated query operator, `QUERY_PRIMARY_KEY_LOOKUP`, for eligible ordinary
`MATCH` clauses.

- The planner recognizes a single-node, single-table, relationship-free `MATCH` with one equality
  predicate normalized to `node.primary_key = upstream_expression`.
- The RHS must be evaluable from the existing child schema and depend on an upstream expression.
- The target table must have a persistent primary-key index. Otherwise, planning falls back to the
  existing scan/filter path.
- The new lookup is chained directly onto the upstream plan. Consecutive eligible `MATCH` clauses
  therefore share the same `UNWIND` input rather than creating independent scans.
- The physical operator evaluates keys batch by batch, probes `NodeTable::lookupPK()`, materializes
  the matched node ID and requested properties, and removes null or missing keys from the batch
  selection vector.
- Non-PK predicates remain as residual filters after the lookup.

The implementation is deliberately separate from COPY's `IndexLookup`, preserving COPY error
handling and its existing no-index fallback behavior.

## Result

For eligible row-driven primary-key matches, the plan changes from a full node-table scan plus
filter to primary-key probes over the upstream rows.

- Query work changes from processing scan/cross-product candidates to one PK probe per input row,
  plus property reads for matched nodes.
- Operator-owned lookup memory is bounded by the vectorized execution batch and requested output
  columns, rather than by target table cardinality.
- Missing or null keys follow ordinary `MATCH` semantics: the corresponding input row is dropped
  without an error or warning.
- Existing constant/literal/parameter PK predicates continue to use
  `PRIMARY_KEY_SCAN_NODE_TABLE`.
- Relationship `COPY FROM` is unchanged. A `MERGE` following the new lookup uses the normal
  transactional relationship-write path and does not use COPY's edge partitioning buffers.

## Scope and Compatibility

The rewrite is intentionally limited to ordinary `MATCH` and does not apply to:

- `OPTIONAL MATCH` or MERGE's internal optional-match planning path;
- multi-table or multi-label patterns, including ANY-graph label patterns;
- relationship patterns;
- `IN`, range predicates, composite keys, secondary indexes, or multiple keys per input row;
- tables without a persistent PK index.

When a condition is not met, the planner retains the existing behavior.

## Validation

The added API coverage verifies:

- `EXPLAIN` contains `QUERY_PRIMARY_KEY_LOOKUP` for a two-MATCH `UNWIND` plus `MERGE` workload;
- rows with a missing endpoint do not create a relationship;
- residual predicates remain effective after the lookup;
- tables without a persistent PK index fall back to `SCAN_NODE_TABLE` and preserve results.

The changed C/C++ files have also been checked with the repository's `clang-format-18` wrapper.


ericyuanhui:
@adsharma any comments？

adsharma:
The idea makes sense to me. Have been tied up with a bunch of other tasks. Will get to it in the next couple of days. Thank you for the submission!

src/include/planner/operator/logical_operator.h
```diff
@@ -43,6 +43,7 @@ enum class LogicalOperatorType : uint8_t {
     HASH_JOIN,
     IMPORT_DATABASE,
     INDEX_LOOK_UP,
+    QUERY_PRIMARY_KEY_LOOKUP,
```
adsharma: Alphabetical order?

adsharma:
### 2. `outputGroupPos` is taken from the first dependent only

`src/planner/plan/plan_subquery.cpp`:

```cpp
const auto outputGroupPos =
    plan.getSchema()->getGroupPos(*getDependentExprs(key, *plan.getSchema())[0]);
```

The first dependent of `key` is used to decide which group `nodeID` and properties land in. This is fine for the common UNWIND case (all the dependent expressions are in the same flat group), but it silently assumes a single group. If a future case has the key referencing expressions in two groups, this would silently put the new outputs in the wrong group. A small comment noting the single-group assumption, or a `KU_DASSERT` that all dependents share a group, would make this safer.

### 3. Cardinality is set to the input cardinality (upper bound)

```cpp
lookup->setCardinality(plan.getCardinality());
```

For a `UNWIND RANGE(1, 1000000) AS id MATCH (n:L {id: id})` where only 1000 rows actually exist, the operator's true output is ~1000 rows but the plan reports 1,000,000. Downstream optimizers (e.g. join ordering, hash-table sizing for the parent join) may over-allocate. For a leaf this is usually fine, but a `min(input, tableSize)` or an estimate based on the table's PK index selectivity would be more honest. The existing code in `factorization_rewriter.cpp` already uses `cardinalityEstimator` (e.g. for `LogicalFlatten`'s cardinality) — consider using it here too.

### 4. Multiple PK matches are not handled

If a `MATCH` has `WHERE n.pk1 = $x AND n.pk2 = $y` (a composite-PK lookup), only the first one wins as the `key`; the rest fall through to residual filters, which then do column scans / index lookups anyway. The new code only handles the single-column-PK case. That's a reasonable scope for this change, but it's worth a comment or TODO so the next person doesn't think it was overlooked.

### 5. `tryPlanQueryPrimaryKeyLookup` is not wired up in subquery / optional-match paths

`planRegularMatch` calls `tryPlanQueryPrimaryKeyLookup` first, but `planSubquery` and `planOptionalMatch` do not. So:

```cypher
MATCH (a) WHERE EXISTS {
  MATCH (b:T) WHERE b.pk = a.fk
} RETURN a
```

…will not benefit from this optimization, even though it's the same shape. That's fine for the first cut (the EXISTS form goes through `SubqueryPredicatePullUpAnalyzer` and becomes an inner join), but it would be worth a follow-up to also rewrite correlated PK lookups. No action needed for this commit, but worth a comment near the call site.

### 6. `LookupNoIndex` test depends on a runtime setting

`UnwindQueryPrimaryKeyLookupFallsBackWithoutIndex` uses `CALL enable_default_hash_index=false;` to make `tryGetPrimaryKeyIndex()` return `nullptr`. This setting is real and works, but:

- If the test ever runs in a fixture where the connection is shared, the setting leaks. The `ApiTest` fixture looks like it creates a fresh connection per test, so this is probably fine — but it's a fragile pattern.
- The test only verifies the planner's branch, not the storage-level path. A more direct unit test of `tryPlanQueryPrimaryKeyLookup` returning false on `tryGetPrimaryKeyIndex() == nullptr` would be more targeted.

Minor; the test passes and exercises the right thing.

### 7. `QueryPrimaryKeyLookup` doesn't declare a print info

`map_query_primary_key_lookup.cpp` constructs `OPPrintInfo` with no fields:

```cpp
auto printInfo = std::make_unique<OPPrintInfo>();
```

So `EXPLAIN` will print the operator type but no useful detail (no table name, no key expression, no properties). Compare to `ScanNodeTablePrintInfo` which is a full struct with tableNames, alias, and properties. For debuggability, a small `QueryPrimaryKeyLookupPrintInfo` (tableName + key-toString + properties) would be helpful. Not a correctness issue.

adsharma:
For 2 and 4, you could skip the optimization instead of `DASSERT`.
5 could be a new issue to track potential future optimizations.

src/include/planner/operator/logical_operator.h
ericyuanhui: fix

ericyuanhui:
> For 2 and 4, you could skip the optimization instead of `DASSERT`. 5 could be a new issue to track potential future optimizations.

For 4. Composite PK matches
This is not applicable to the current implementation because Ladybug currently supports only a single-column primary key. The DDL/catalog path stores one primary-key property name, and the binder validates a single property. Supporting composite primary keys would require a broader change across DDL, index representation, lookup APIs, and planner/physical operators, rather than an adjustment to this optimization alone.

For 6. LookupNoIndex test
The test is intentionally an integration test: it verifies both that the planner does not select QUERY_PRIMARY_KEY_LOOKUP without a primary-key index and that the fallback query still returns the correct result.
The setting does not leak across tests in the current fixture setup. Each ApiTest creates a fresh database and connection per test, and enable_default_hash_index is applied before the test table is created. A narrower unit test around tryPlanQueryPrimaryKeyLookup would require planner/storage setup or mocks, while the existing test exercises the relevant behavior end to end.

ericyuanhui:
> For 2 and 4, you could skip the optimization instead of `DASSERT`. 5 could be a new issue to track potential future optimizations.

For 5 I add the comment at the call site first.

ericyuanhui:
> ### 2. `outputGroupPos` is taken from the first dependent only
> `src/planner/plan/plan_subquery.cpp`:
> 
> ```c++
> const auto outputGroupPos =
>     plan.getSchema()->getGroupPos(*getDependentExprs(key, *plan.getSchema())[0]);
> ```
> 
> The first dependent of `key` is used to decide which group `nodeID` and properties land in. This is fine for the common UNWIND case (all the dependent expressions are in the same flat group), but it silently assumes a single group. If a future case has the key referencing expressions in two groups, this would silently put the new outputs in the wrong group. A small comment noting the single-group assumption, or a `KU_DASSERT` that all dependents share a group, would make this safer.
> 
> ### 3. Cardinality is set to the input cardinality (upper bound)
> ```c++
> lookup->setCardinality(plan.getCardinality());
> ```
> 
> For a `UNWIND RANGE(1, 1000000) AS id MATCH (n:L {id: id})` where only 1000 rows actually exist, the operator's true output is ~1000 rows but the plan reports 1,000,000. Downstream optimizers (e.g. join ordering, hash-table sizing for the parent join) may over-allocate. For a leaf this is usually fine, but a `min(input, tableSize)` or an estimate based on the table's PK index selectivity would be more honest. The existing code in `factorization_rewriter.cpp` already uses `cardinalityEstimator` (e.g. for `LogicalFlatten`'s cardinality) — consider using it here too.
> 
> ### 4. Multiple PK matches are not handled
> If a `MATCH` has `WHERE n.pk1 = $x AND n.pk2 = $y` (a composite-PK lookup), only the first one wins as the `key`; the rest fall through to residual filters, which then do column scans / index lookups anyway. The new code only handles the single-column-PK case. That's a reasonable scope for this change, but it's worth a comment or TODO so the next person doesn't think it was overlooked.
> 
> ### 5. `tryPlanQueryPrimaryKeyLookup` is not wired up in subquery / optional-match paths
> `planRegularMatch` calls `tryPlanQueryPrimaryKeyLookup` first, but `planSubquery` and `planOptionalMatch` do not. So:
> 
> ```cypher
> MATCH (a) WHERE EXISTS {
>   MATCH (b:T) WHERE b.pk = a.fk
> } RETURN a
> ```
> 
> …will not benefit from this optimization, even though it's the same shape. That's fine for the first cut (the EXISTS form goes through `SubqueryPredicatePullUpAnalyzer` and becomes an inner join), but it would be worth a follow-up to also rewrite correlated PK lookups. No action needed for this commit, but worth a comment near the call site.
> 
> ### 6. `LookupNoIndex` test depends on a runtime setting
> `UnwindQueryPrimaryKeyLookupFallsBackWithoutIndex` uses `CALL enable_default_hash_index=false;` to make `tryGetPrimaryKeyIndex()` return `nullptr`. This setting is real and works, but:
> 
> * If the test ever runs in a fixture where the connection is shared, the setting leaks. The `ApiTest` fixture looks like it creates a fresh connection per test, so this is probably fine — but it's a fragile pattern.
> * The test only verifies the planner's branch, not the storage-level path. A more direct unit test of `tryPlanQueryPrimaryKeyLookup` returning false on `tryGetPrimaryKeyIndex() == nullptr` would be more targeted.
> 
> Minor; the test passes and exercises the right thing.
> 
> ### 7. `QueryPrimaryKeyLookup` doesn't declare a print info
> `map_query_primary_key_lookup.cpp` constructs `OPPrintInfo` with no fields:
> 
> ```c++
> auto printInfo = std::make_unique<OPPrintInfo>();
> ```
> 
> So `EXPLAIN` will print the operator type but no useful detail (no table name, no key expression, no properties). Compare to `ScanNodeTablePrintInfo` which is a full struct with tableNames, alias, and properties. For debuggability, a small `QueryPrimaryKeyLookupPrintInfo` (tableName + key-toString + properties) would be helpful. Not a correctness issue.

For 3 I will fix later. I should find out the right way. 
`CardinalityEstimator` currently has table cardinality, index availability, and limited column statistics, but it does not have statistics about the overlap between the input key distribution and the target table's primary-key domain. Primary-key uniqueness only guarantees that each input key can match at most one row; it does not indicate how many input keys will actually match.

ericyuanhui:
For 7 I have add QueryPrimaryKeyLookupPrintInfo to explain.

PR #729, opened 2026-07-23
fix: add SQL-standard '' single quote escape in Cypher string literals

chiangchenghsin-hash:
## Summary

Adds SQL-92 standard two-single-quotes (`''`) escape support within Cypher single-quoted string literals, complementing the existing backslash (`\'`) escape.

## Problem

Cypher uses backslash escapes for strings (`'It\'s a test'`), which differs from the SQL standard (`'It''s a test'`). This causes parser exceptions when users write SQL-style escaped strings — a common pitfall for those coming from SQL backgrounds.

## Fix

Two changes:

1. **`src/parser/transformer.cpp`**: In `transformStringLiteral()`, add handling for consecutive single quotes (`''`) before the fallback `else` branch — translates `''` to `'`.

2. **`src/antlr4/Cypher.g4`**: Add `'\'\''` as an alternative within the single-quoted `StringLiteral` rule, so the ANTLR lexer recognizes `''` as a valid token sequence inside a string literal.

## Verification

- Windows 11 x64, LadybugDB 0.18.3
- `"It's working"` (double-quote workaround for single quotes) — ✅ already works
- `'It''s a test'` (SQL-style escape) — ✅ fixed (was Parser exception before)


prrao87:
Hi Arun, any reason this was closed? Seems like a good addition.

PR #733, opened 2026-07-24
fix: restore/save selection vector in QueryPrimaryKeyLookup

adsharma:
Row-driven primary-key lookup (e92346c) filtered the node-id vector's shared DataChunkState directly. When an UNWIND row missed the PK index (e.g. UNWIND [0, 2, 3] ... MATCH ({id: x}) with x=0 absent), the selection size was driven to 0 and the upstream Flatten/Unwind operator kept seeing an empty selection, starving every subsequent tuple so only 0 rows were returned despite later keys (2, 3) existing.

Inherit SelVectorOverWriter (like Filter/NodeLabelFilter) and wrap each child pull with restore/save so we rewrite a private selection vector instead of clobbering the child's. Skip-empty now loops back to the child instead of terminating the pipeline.

Adds regression test UnwindQueryPrimaryKeyLookupLeadingMissThenHits.

### DIFF
--- a/src/include/processor/operator/query_primary_key_lookup.h
+++ b/src/include/processor/operator/query_primary_key_lookup.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "expression_evaluator/expression_evaluator.h"
+#include "processor/operator/filtering_operator.h"
 #include "processor/operator/physical_operator.h"
 #include "processor/operator/scan/scan_node_table.h"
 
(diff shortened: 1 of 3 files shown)


adsharma:
Cc: @ericyuanhui 

```
CREATE NODE TABLE t(id INT64, v INT64, PRIMARY KEY(id));
CREATE (t2:t {id: 2, v: 20});
CREATE (t3:t {id: 3, v: 30});
UNWIND [0, 2, 3] as x MATCH (t:t {id: x}) RETURN t.v;
```

ericyuanhui:
> UNWIND [0, 2, 3] as x MATCH (t:t {id: x}) RETURN t.v;

yes . I get the wrong path. Thank you for your help!

Issue #678, opened 2026-07-11
Bug: Partial string scans corrupt checkpointed NODE properties after DETACH DELETE (node-table twin of #658)


hsinister:
### Ladybug version

0.18.1 (Python wheel). This release already contains the #658 fix (PR #659, merged 2026-07-09; release tagged 2026-07-10). `main` is currently at the exact same commit as the `v0.18.1` tag (`1354081eb5528b3ca12e38dd4402cdd47215e57a`), so this is not fixed on `main`.

### What operating system are you using?

Linux x86_64 (Ubuntu 24.04.4 LTS), Python 3.12. Not platform specific — the failing path is in storage checkpoint/scan.

### What happened?

`#658` fixed partial-string-scan corruption for **relationship** properties. The **node** property path has the same class of bug and is still broken in 0.18.1: deleting one or more **node** rows from a table that has an out-of-line `STRING` property, followed by a checkpoint, corrupts the **sequential-scan** read path of that string property for *surviving* rows after reload. The primary-key point lookup returns the correct value; only the scan projection is wrong. The divergence can be severe — in the synthetic repro below, multiple surviving rows per run read wrong, and a scan can return non-UTF-8 bytes.

Concretely, for surviving nodes after a delete + checkpoint + reopen:

- `MATCH (n:N) RETURN n.txt` (sequential scan) returns a wrong/garbled value, and
- `MATCH (n:N {id: $id}) RETURN n.txt` (PK point lookup) returns the correct value.

The delete itself is correct (count drops as expected, endpoints/ids stable); only the projected string property drifts, silently, at the query layer — the same failure shape #658 describes, but on node tables.

We independently bisected this against #658's fix boundary and confirmed:

- relationship-only `DELETE` + checkpoint is **safe** on 0.18.1 (consistent with #658 being fixed), and
- node `DETACH DELETE` + checkpoint **corrupts** the node string scan path.

### Mechanism hypothesis (ties directly to #658)

We believe this is the same coordinate-system mismatch #658 identified, reached through the node-table checkpoint path rather than the relationship one.

[...]

PR #737, opened 2026-07-27
fix: bound filtered string scans to the current segment

kory-io:
## Description

Addresses the non-concurrent string scan corruption reported in #678.

`StringColumn::scanFiltered` receives selection positions in result-vector
coordinates. For a scan spanning multiple persisted segments, it checked the
lower result offset but did not bound selected positions to the current
segment's result range. A position belonging to a later segment could therefore
be read from the current segment using the wrong local offset.

Pass the current segment's scan length into `scanFiltered`, restrict selected
positions to `[offsetInResult, offsetInResult + numValuesToScan)`, and derive
the index-column offset from the segment-local coordinate.

The regression test constructs two persisted string segments and selects only
result positions owned by the second segment. It verifies that the first
segment does not consume those positions and that the second segment returns
the expected strings.

This changes only scan bounds and does not change the on-disk format.

## Validation

- `build/release/test/storage/string_chunk_scan_test --gtest_color=no`
- `ctest --test-dir build/release/test --output-on-failure -j 4`
- `python3 scripts/run-clang-format.py --clang-format-executable clang-format-18 ...`

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation Update

## Checklist

- [x] I have changed storage version if on disk format has changed. — N/A; no on-disk format change.
- [ ] I have requested a review from a maintainer.
- [x] I have updated the documentation (if needed). — N/A; no user-facing API or configuration change.

### DIFF
--- a/src/include/storage/table/string_column.h
+++ b/src/include/storage/table/string_column.h
@@ -44,7 +44,8 @@ protected:
         common::offset_t numValuesToRead, common::ValueVector* resultVector,
         common::sel_t startPosInVector = 0) const;
     void scanFiltered(const SegmentState& state, common::offset_t startOffsetInChunk,
-        common::ValueVector* resultVector, common::sel_t startPosInVector) const;
+        common::row_idx_t numValuesToScan, common::ValueVector* resultVector,
+        common::sel_t startPosInVector) const;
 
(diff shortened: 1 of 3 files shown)


adsharma:
Thank you for fixing!

PR #741, opened 2026-07-27
Fix/foreign db name dynamic lookup

adsharma:
These fixes are needed for node + fkrel table pushdown to an attached SQL db to work properly.


adsharma:
This doesn't work

PR #765, opened 2026-07-31
Trivial query overhead2

adsharma:
temp pr to run ci for #764


adsharma:
This will land as #764

PR #768, opened 2026-08-01
Fix #757: accept FixedSizeBinary(16) with arrow.uuid for UUID PK

### LINKED ISSUE #757: Bug: createArrowRelTable cannot address a UUID-keyed node table, though COPY can
I hit this staging relationships through Arrow for a schema whose node tables are keyed on `UUID`. Verified against `lbug` 0.19.0 and a `relwithdebinfo` build of `3abdb0653`; the reproducer below is plain C against `c_api/lbug.h`, so no binding is involved.

## What happens

`createArrowRelTable` rejects every endpoint encoding when the endpoint node table has a `UUID` primary key:

| endpoint encoding | result | message |
|---|---|---|
| `Utf8` | **fails** | `Arrow 'from' column type STRING must match source node PK type UUID` |
| `LargeUtf8` | **fails** | `Arrow 'from' column type STRING must match source node PK type UUID` |
| `Binary` | **fails** | `Arrow 'from' column type BLOB must match source node PK type UUID` |
| `FixedSizeBinary(16)` | **fails** | `Arrow 'from' column type BLOB must match source node PK type UUID` |

The identical call against a `STRING`-keyed node table succeeds. There is no encoding that maps to `UUID`, so the Arrow relationship path is unavailable for any schema keyed on `UUID`.

## The same conversion is accepted everywhere else

| path | result |
|---|---|
| CSV `STRING` column into a `UUID` primary key | ok |
| CSV `STRING` endpoints into a `UUID`-keyed rel table | ok |
| stage the edge list as an Arrow *node* table, then `COPY` with a cast | ok |

The second row is the same operation as the failing one, on the same rel table shape, through a different door: `COPY R FROM 'edges.csv'` resolves string endpoints against a `UUID` primary key without complaint. Only `createArrowRelTable` refuses.

## Why

`ArrowRelTable`'s constructor compares type *strings* for exact equality, `src/storage/table/arrow_rel_table.cpp:111`:

```cpp
if (srcArrowType.toString() != srcPKType.toString()) {
    throw RuntimeException("Arrow 'from' column type " + srcArrowType.toString() +
                           " must match source node PK type " + srcPKType.toString());
}
```

An Arrow `Utf8` maps to `STRING`, the PK is `UUID`, and `"STRING" != "UUID"`.

## Why I am not sending a patch

Relaxing the predicate on its own would be wrong. The endpoint key vector is allocated with the *primary key's* logical type (`:214`):

```cpp
auto srcPKType = fromNodeTable->getColumn(fromNodeTable->getPKColumnID()).getDataType().copy();
relScanState.arrowSrcKeyVector =
    std::make_unique<ValueVector>(std::move(srcPKType), memoryManager, singleValueState);
```

and the endpoint value is then read straight out of the Arrow buffer into it (`:292`, via `readSingleArrowValue` at `:228`, which calls `ArrowConverter::fromArrowArray` with no conversion step). With a `Utf8` array and a `UUID` vector that writes string data into a 16-byte fixed-width vector. The strict check is guarding a real limitation, not being pedantic, so the fix is a conversion in the scan path rather than a looser comparison, and where that conversion belongs is your call.

Options as I see them, in the order I would want them:

1. **Convert on read.** Accept an implicitly-castable endpoint type and cast into the PK-typed vector in `readSingleArrowValue`, keeping the existing message for types that genuinely cannot be reconciled. Most useful, most work.
2. **Accept `FixedSizeBinary(16)` for `UUID` specifically.** The physical layouts already match, so this is a type-mapping entry rather than a conversion, and it covers the case that motivated this report. It does mean a caller has to know the byte order you expect.
3. **Leave it and document it**, saying plainly that Arrow relationship tables require an exact PK type match and listing which Arrow types map to which. That is worth doing even if you do 1 or 2, because the current message reads as "you passed the wrong thing" when in fact nothing can be passed.

## Workaround, for anyone else hitting this

Stage the edge list as a *node* table and let the `COPY` subquery resolve the endpoints:

```
COPY R FROM (MATCH (e:stg), (a:X), (b:Y)
             WHERE a.id = cast(e.from AS UUID) AND b.id = cast(e.to AS UUID)
             RETURN a.id, b.id, e.position) (from='X', to='Y');
```

Verified in the reproducer: it carries edge properties and reads back correctly. For a rel table declared over more than one node-table pair, the `COPY` needs the explicit `(from=…, to=…)` options as usual.

`createArrowRelTableCSR` is unaffected, since it addresses nodes by internal offset rather than by key. It was not a practical substitute in my case for unrelated performance reasons, so I did not pursue it.

## Tests

`test/api/arrow_rel_table_test.cpp` is the file to extend. `ArrowRelTableTest.ScanArrowRelTableOverNativeNodeTable` is the closest existing case: the same with `UUID`-keyed endpoint tables and `Utf8` from/to columns, asserting the relationships load and read back. Worth adding whichever option you take, since option 3 also has an assertable outcome (a message that says the exact-match rule is by design).

## Reproducer

```
cc -o arrow_rel_pk arrow_rel_pk.c -I$LBUG/src/include -I$LBUG/src/include/c_api \
   -L$LBUG/build/relwithdebinfo/src -llbug -Wl,-rpath,$LBUG/build/relwithdebinfo/src
./arrow_rel_pk
```

<details>
<summary><code>arrow_rel_pk.c</code></summary>

```c
/* arrow_rel_pk.c - can createArrowRelTable address a UUID-keyed node table?
 *
 * cc -o arrow_rel_pk arrow_rel_pk.c -I$LBUG/src/include -I$LBUG/src/include/c_api \
 *    -L$LBUG/build/relwithdebinfo/src -llbug -Wl,-rpath,$LBUG/build/relwithdebinfo/src
 *
 * Builds the Arrow C Data Interface structs by hand, so no language binding is involved.
 * The batch is one edge, U1 -> U2, with a `position` property.
 */
#include <lbug.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#define U1 "11111111-1111-1111-1111-111111111111"
#define U2 "22222222-2222-2222-2222-222222222222"

/* ---- minimal Arrow C Data Interface producers ------------------------------ */

static void release_schema(struct ArrowSchema* s) {
    for (int64_t i = 0; i < s->n_children; i++)
        if (s->children[i]->release)
            s->children[i]->release(s->children[i]);
    free(s->children);
    s->release = NULL;
}

static void release_array(struct ArrowArray* a) {
    for (int64_t i = 0; i < a->n_children; i++)
        if (a->children[i]->release)
            a->children[i]->release(a->children[i]);
    free(a->children);
    free((void*)a->buffers);
    a->release = NULL;
}

static void init_schema(struct ArrowSchema* s, const char* fmt, const char* name) {
    memset(s, 0, sizeof *s);
    s->format = fmt;
    s->name = name;
    s->release = release_schema;
}

/* one variable-width string; fmt is "u" (Utf8), "U" (LargeUtf8) or "z" (Binary) */
static void varlen_array(struct ArrowArray* a, const char* fmt, const void* data, size_t len) {
    memset(a, 0, sizeof *a);
    a->length = 1;
    a->null_count = 0;
    a->n_buffers = 3;
    const void** bufs = calloc(3, sizeof(void*));
    bufs[0] = NULL;
    if (fmt[0] == 'U') {
        int64_t* o = calloc(2, sizeof(int64_t));
        o[0] = 0;
        o[1] = (int64_t)len;
        bufs[1] = o;
    } else {
        int32_t* o = calloc(2, sizeof(int32_t));
        o[0] = 0;
        o[1] = (int32_t)len;
        bufs[1] = o;
    }
    bufs[2] = data;
    a->buffers = bufs;
    a->release = release_array;
}

/* one fixed-size binary value, fmt "w:16" */
static void fixed_array(struct ArrowArray* a, const void* data) {
    memset(a, 0, sizeof *a);
    a->length = 1;
    a->null_count = 0;
    a->n_buffers = 2;
    const void** bufs = calloc(2, sizeof(void*));
    bufs[0] = NULL;
    bufs[1] = data;
    a->buffers = bufs;
    a->release = release_array;
}

/* one int64 */
static void int64_array(struct ArrowArray* a, int64_t* slot, int64_t v) {
    *slot = v;
    memset(a, 0, sizeof *a);
    a->length = 1;
    a->null_count = 0;
    a->n_buffers = 2;
    const void** bufs = calloc(2, sizeof(void*));
    bufs[0] = NULL;
    bufs[1] = slot;
    a->buffers = bufs;
    a->release = release_array;
}

static void uuid_bytes(const char* text, unsigned char out[16]) {
    int j = 0;
    for (int i = 0; text[i] && j < 16; i++) {
        if (text[i] == '-')
            continue;
        unsigned hi, lo;
        sscanf(text + i, "%1x", &hi);
        sscanf(text + i + 1, "%1x", &lo);
        out[j++] = (unsigned char)((hi << 4) | lo);
        i++;
    }
}

/* from/to in `fmt`, plus an INT64 `position` */
static void build_rel(struct ArrowSchema* schema, struct ArrowArray* array, const char* fmt,
    void* fromBuf, void* toBuf, size_t len, int64_t* posSlot) {
    init_schema(schema, "+s", NULL);
    schema->n_children = 3;
    schema->children = calloc(3, sizeof(struct ArrowSchema*));
    const char* names[3] = {"from", "to", "position"};
    const char* fmts[3] = {fmt, fmt, "l"};
    for (int i = 0; i < 3; i++) {
        schema->children[i] = calloc(1, sizeof(struct ArrowSchema));
        init_schema(schema->children[i], fmts[i], names[i]);
    }

    memset(array, 0, sizeof *array);
    array->length = 1;
    array->n_buffers = 1;
    const void** rb = calloc(1, sizeof(void*));
    rb[0] = NULL;
    array->buffers = rb;
    array->n_children = 3;
    array->children = calloc(3, sizeof(struct ArrowArray*));
    for (int i = 0; i < 3; i++)
        array->children[i] = calloc(1, sizeof(struct ArrowArray));
    if (!strcmp(fmt, "w:16")) {
        fixed_array(array->children[0], fromBuf);
        fixed_array(array->children[1], toBuf);
    } else {
        varlen_array(array->children[0], fmt, fromBuf, len);
        varlen_array(array->children[1], fmt, toBuf, len);
    }
    int64_array(array->children[2], posSlot, 7);
    array->release = release_array;
}

/* ---- helpers --------------------------------------------------------------- */

static int run(lbug_connection* c, const char* q, char* msg, size_t cap) {
    lbug_query_result r;
    memset(&r, 0, sizeof r);
    lbug_state s = lbug_connection_query(c, q, &r);
    int ok = (s == LbugSuccess) && lbug_query_result_is_success(&r);
    if (msg)
        msg[0] = '\0';
    if (!ok && msg) {
        char* m = lbug_query_result_get_error_message(&r);
        snprintf(msg, cap, "%s", m && *m ? m : "(no message)");
        if (m)
            free(m);
    }
    lbug_query_result_destroy(&r);
    return ok;
}

static void must(lbug_connection* c, const char* q) {
    char msg[512];
    if (!run(c, q, msg, sizeof msg)) {
        fprintf(stderr, "setup failed: %s\n  -> %s\n", q, msg);
        exit(1);
    }
}

static int try_rel(lbug_connection* conn, const char* relName, const char* src, const char* dst,
    const char* fmt, void* fromBuf, void* toBuf, size_t len, char* msg, size_t cap) {
    struct ArrowSchema schema;
    struct ArrowArray array;
    int64_t pos;
    build_rel(&schema, &array, fmt, fromBuf, toBuf, len, &pos);

    lbug_query_result out;
    memset(&out, 0, sizeof out);
    lbug_state st = lbug_connection_create_arrow_rel_table(conn, relName, src, dst, &schema, &array,
        1, &out);
    int ok = (st == LbugSuccess) && lbug_query_result_is_success(&out);
    msg[0] = '\0';
    if (!ok) {
        char* m = lbug_query_result_get_error_message(&out);
        if (m && *m) {
            snprintf(msg, cap, "%s", m);
            free(m);
        } else {
            char* le = lbug_get_last_error();
            snprintf(msg, cap, "%s", le && *le ? le : "(no message)");
            if (le)
                free(le);
        }
    }
    lbug_query_result_destroy(&out);
    return ok;
}

int main(void) {
    lbug_database db;
    lbug_connection conn;
    if (lbug_database_init("", lbug_default_system_config(), &db) != LbugSuccess) {
        fprintf(stderr, "database_init failed\n");
        return 1;
    }
    lbug_connection_init(&db, &conn);

    must(&conn, "CREATE NODE TABLE NU(id UUID, name STRING, PRIMARY KEY(id));");
    must(&conn, "CREATE NODE TABLE NS(id STRING, name STRING, PRIMARY KEY(id));");
    must(&conn, "CREATE (:NU {id: UUID('" U1 "'), name:'a'});");
    must(&conn, "CREATE (:NU {id: UUID('" U2 "'), name:'b'});");
    must(&conn, "CREATE (:NS {id: '" U1 "', name:'a'});");
    must(&conn, "CREATE (:NS {id: '" U2 "', name:'b'});");

    char u1[] = U1, u2[] = U2;
    unsigned char b1[16], b2[16];
    uuid_bytes(U1, b1);
    uuid_bytes(U2, b2);
    char msg[600];

    printf("### createArrowRelTable against a UUID-keyed node table\n\n");
    printf("| endpoint encoding | result | message |\n|---|---|---|\n");
    struct {
        const char* label;
        const char* fmt;
        void *f, *t;
        size_t len;
    } cases[] = {
        {"Utf8", "u", u1, u2, 36},
        {"LargeUtf8", "U", u1, u2, 36},
        {"Binary", "z", b1, b2, 16},
        {"FixedSizeBinary(16)", "w:16", b1, b2, 16},
        {NULL, NULL, NULL, NULL, 0},
    };
    for (int i = 0; cases[i].label; i++) {
        char rel[32];
        snprintf(rel, sizeof rel, "RU%d", i);
        int ok = try_rel(&conn, rel, "NU", "NU", cases[i].fmt, cases[i].f, cases[i].t,
            cases[i].len, msg, sizeof msg);
        printf("| %s | %s | %s |\n", cases[i].label, ok ? "ok" : "**fails**", ok ? "" : msg);
    }

    printf("\n### the same call against a STRING-keyed node table (control)\n\n");
    printf("| endpoint encoding | result | message |\n|---|---|---|\n");
    int ok = try_rel(&conn, "RS0", "NS", "NS", "u", u1, u2, 36, msg, sizeof msg);
    printf("| Utf8 | %s | %s |\n", ok ? "ok" : "**fails**", ok ? "" : msg);

    printf("\n### the same conversion through neighbouring paths\n\n");
    printf("| path | result | message |\n|---|---|---|\n");

    FILE* f = fopen("/tmp/arrow_rel_pk_nodes.csv", "w");
    fprintf(f, "id,name\n" U1 ",a\n" U2 ",b\n");
    fclose(f);
    f = fopen("/tmp/arrow_rel_pk_edges.csv", "w");
    fprintf(f, "from,to,position\n" U1 "," U2 ",7\n");
    fclose(f);

    must(&conn, "CREATE NODE TABLE NU3(id UUID, name STRING, PRIMARY KEY(id));");
    ok = run(&conn, "COPY NU3 FROM '/tmp/arrow_rel_pk_nodes.csv' (HEADER=true);", msg, sizeof msg);
    printf("| CSV `STRING` column into a `UUID` primary key | %s | %s |\n", ok ? "ok" : "**fails**",
        ok ? "" : msg);

    must(&conn, "CREATE REL TABLE RU_CSV(FROM NU3 TO NU3, position INT64);");
    ok = run(&conn, "COPY RU_CSV FROM '/tmp/arrow_rel_pk_edges.csv' (HEADER=true);", msg,
        sizeof msg);
    printf("| CSV `STRING` endpoints into a `UUID`-keyed rel table | %s | %s |\n",
        ok ? "ok" : "**fails**", ok ? "" : msg);

    /* the workaround: stage the edge list as an Arrow NODE table, resolve in the subquery */
    {
        struct ArrowSchema schema;
        struct ArrowArray array;
        int64_t pos;
        build_rel(&schema, &array, "u", u1, u2, 36, &pos);
        lbug_query_result out;
        memset(&out, 0, sizeof out);
        lbug_state st =
            lbug_connection_create_arrow_table(&conn, "stge", &schema, &array, 1, &out);
        int created = (st == LbugSuccess) && lbug_query_result_is_success(&out);
        lbug_query_result_destroy(&out);
        if (created) {
            must(&conn, "CREATE REL TABLE RU_STG(FROM NU TO NU, position INT64);");
            ok = run(&conn,
                "COPY RU_STG FROM (MATCH (e:stge), (a:NU), (b:NU) "
                "WHERE a.id = cast(e.from AS UUID) AND b.id = cast(e.to AS UUID) "
                "RETURN a.id, b.id, e.position);",
                msg, sizeof msg);
        } else {
            ok = 0;
            snprintf(msg, sizeof msg, "could not stage the edge list");
        }
        printf("| workaround: stage edges as an Arrow node table, then `COPY` | %s | %s |\n",
            ok ? "ok" : "**fails**", ok ? "" : msg);
        if (ok) {
            lbug_query_result r;
            memset(&r, 0, sizeof r);
            if (lbug_connection_query(&conn, "MATCH (a:NU)-[r:RU_STG]->(b:NU) RETURN count(*);",
                    &r) == LbugSuccess) {
                lbug_flat_tuple ft;
                if (lbug_query_result_get_next(&r, &ft) == LbugSuccess) {
                    lbug_value v;
                    if (lbug_flat_tuple_get_value(&ft, 0, &v) == LbugSuccess) {
                        int64_t n = 0;
                        lbug_value_get_int64(&v, &n);
                        printf("\nworkaround row count: %lld\n", (long long)n);
                        lbug_value_destroy(&v);
                    }
                    lbug_flat_tuple_destroy(&ft);
                }
            }
            lbug_query_result_destroy(&r);
        }
    }

    lbug_connection_destroy(&conn);
    lbug_database_destroy(&db);
    return 0;
}
```

</details>

## Notes

Verified against `lbug` 0.19.0 and a `relwithdebinfo` build of `3abdb0653`. `src/storage/table/arrow_rel_table.cpp` has had one commit since the 0.18.2 tag, where I first saw this (`94ebf1b14`, CSR degree counts); `git log -L` over the validation block shows it untouched.

adsharma:
## Summary

Closes #757.

`createArrowRelTable` required an exact match between the Arrow endpoint
column's logical type and the node table's primary key type. No Arrow wire
type mapped to a UUID logical type, so `createArrowRelTable` was unable to
address any node table keyed on UUID. CSV `COPY` could resolve STRING
endpoints against a UUID PK, so the restriction was an asymmetry, not a
limitation of the storage layer.

This implements **option 2** from the issue: accept
`FixedSizeBinary(16)` carrying the `arrow.uuid` Arrow extension as a
UUID primary key. The 16-byte physical layout is identical to lbug's
internal UUID storage, so no conversion is needed in the read path beyond
the byte order and the MSB flip that UUIDs already require internally.

## Changes

- Recognize the `arrow.uuid` Arrow extension on `FixedSizeBinary(16)` as
  a UUID in `tryGetArrowLogicalTypeInfo`, so the schema round-trips to
  `LogicalType::UUID` instead of `BLOB`.
- Route a `FixedSizeBinary(16)` UUID column through a new
  `scanArrowArrayUuid` that byte-reverses the 16-byte big-endian buffer
  to land in the `int128_t` layout and re-flips the UUID MSB. This mirrors
  the byte order the existing UUID write path produces.
- Replace the exact-type check in the `ArrowRelTable` constructor with
  `isArrowEndpointCompatibleWithPK`, which still accepts an exact
  logical-type match and additionally accepts a `FixedSizeBinary(16)`
  UUID-extension column for a UUID PK. The error message is updated to
  describe the new acceptance rule.
- Test utilities gain `createUuidSchema` and `createUuidArray` helpers.

## Tests

- `ArrowRelTableTest.ScanArrowRelTableOverNativeUuidNodeTable` is the
  end-to-end regression test: a `CREATE NODE TABLE ... PRIMARY KEY(id
  UUID)`, then `createRelTableFromArrowTable` with `w:16`+`arrow.uuid`
  endpoint columns. Asserts the edges load, the weights sum, and the
  endpoints round-trip back to the right nodes.
- `ArrowRelTableTest.RejectFixedSizeBinaryWithoutUuidExtensionForUuidPK`
  asserts that a bare `w:16` (no extension) is still rejected for a UUID
  PK, since the current "exact match" message contract is preserved for
  types that cannot be reconciled.
- `ArrowConverterTest.bindsArrowUuidExtensionMetadataAsUuid`,
  `...fixedSizeBinaryWithoutUuidExtensionIsNotUuid`, and
  `...rejectsNon16ByteArrowUuidExtensionAsUuid` cover the metadata
  parser directly: a correctly-annotated 16-byte column binds to UUID,
  a bare 16-byte column stays BLOB, and a non-16-byte column with the
  UUID extension is rejected (a UUID is exactly 16 bytes).

## Notes

- Byte order on the wire is big-endian (network), matching the Arrow
  UUID extension spec. The 16 bytes written by lbug's own UUID
  serialization in `arrow_row_batch.cpp` are already in this order; the
  read path mirrors the byte reversal performed on write.
- A caller now needs to know the byte order, as the issue notes, and
  must include the `ARROW:extension:name=arrow.uuid` extension metadata
  on the endpoint columns. The new error message spells that out.
- `fromArrowSchema` now returns `UUID` for `w:16` columns carrying the
  extension, so this is also a prerequisite fix for reading UUID values
  from Arrow anywhere else (e.g. node tables).

### DIFF
--- a/src/common/arrow/arrow_array_scan.cpp
+++ b/src/common/arrow/arrow_array_scan.cpp
@@ -280,6 +280,36 @@ static void scanArrowArrayFixedBLOB(const ArrowArray* array, ValueVector& output
     });
 }
 
+// Reads a FixedSizeBinary(16) Arrow column carrying the `arrow.uuid` extension
+// (https://arrow.apache.org/docs/format/CDataInterface.html) into a UUID-typed
+// value vector. The 16 bytes are in big-endian (network) byte order on the wire;
+// lbug stores the UUID internally as an int128 with the most-significant bit
+// flipped (so lexicographic order of strings matches the natural numeric order),
+// so the buffer is byte-reversed to land in the int128 layout, then the MSB is
+// re-flipped.
+static void scanArrowArrayUuid(const ArrowArray* array, ValueVector& outputVector,
+    ArrowNullMaskTree* mask, uint64_t srcOffset, uint64_t dstOffset, uint64_t count) {
+    constexpr uint64_t uuidSize = sizeof(int128_t);
+    auto arrayBuffer = ((const uint8_t*)array->buffers[1]) + srcOffset * uuidSize;
+
+    mask->copyToValueVector(&outputVector, dstOffset, count);
+
+    rowIter(outputVector, count, [&](auto i) {
+        if (mask->isNull(i)) {
+            return;
+        }
+        int128_t value{};
+        auto valueBytes = reinterpret_cast<uint8_t*>(&value);
+        const auto* src = arrayBuffer + i * uuidSize;
+        for (uint64_t j = 0; j < uuidSize; ++j) {
+            valueBytes[j] = src[uuidSize - 1 - j];
+        }
+        // Re-flip the MSB to undo the internal flip applied when writing.
+        value.high ^= (int64_t(1) << 63);
+        outputVector.setValue<int128_t>(i + dstOffset, value);
+    });
+}
+
 template<typename offsetsT>
 static void scanArrowArrayList(const ArrowSchema* schema, const ArrowArray* array,
     ValueVector& outputVector, ArrowNullMaskTree* mask, uint64_t srcOffset, uint64_t dstOffset,
@@ -649,7 +679,11 @@ void ArrowConverter::fromArrowArray(const ArrowSchema* schema, const ArrowArray*
         }
     }
     case 'w':
-        // FIXED BLOB
+        // FIXED BLOB / UUID
+        if (logicalTypeInfo->has_value() &&
+            (*logicalTypeInfo)->type == ArrowLogicalTypeInfo::Type::UUID) {
+            return scanArrowArrayUuid(array, outputVector, mask, srcOffset, dstOffset, count);
+        }
         return scanArrowArrayFixedBLOB(array, outputVector, mask, std::stoi(arrowType + 2),
             srcOffset, dstOffset, count);
     case 't':
--- a/src/common/arrow/arrow_schema_metadata.cpp
+++ b/src/common/arrow/arrow_schema_metadata.cpp
@@ -18,6 +18,10 @@ std::optional<ArrowLogicalTypeInfo> tryGetArrowLogicalTypeInfo(const ArrowSchema
         snowflakeTypeInfo.has_value()) {
         return snowflakeTypeInfo;
     }
+    if (auto uuidTypeInfo = tryParseUuidExtensionMetadata(schema, metadata);
+        uuidTypeInfo.has_value()) {
+        return uuidTypeInfo;
+    }
     return tryParseGenericIntegerBackedDecimalMetadata(schema, metadata);
 }
 
--- a/src/common/arrow/arrow_schema_metadata_internal.h
+++ b/src/common/arrow/arrow_schema_metadata_internal.h
@@ -29,6 +29,8 @@ std::optional<ArrowLogicalTypeInfo> tryParseSnowflakeLogicalTypeInfo(const Arrow
     const ArrowMetadataMap& metadata);
 std::optional<ArrowLogicalTypeInfo> tryParseGenericIntegerBackedDecimalMetadata(
     const ArrowSchema* schema, const ArrowMetadataMap& metadata);
+std::optional<ArrowLogicalTypeInfo> tryParseUuidExtensionMetadata(const ArrowSchema* schema,
+    const ArrowMetadataMap& metadata);
 
 } // namespace common
 } // namespace lbug
--- a/src/common/arrow/arrow_schema_metadata_utils.cpp
+++ b/src/common/arrow/arrow_schema_metadata_utils.cpp
@@ -141,5 +141,23 @@ bool isValidDecimalParameters(uint32_t precision, uint32_t scale) {
     return precision > 0 && precision <= DECIMAL_PRECISION_LIMIT && scale <= precision;
 }
 
+std::optional<ArrowLogicalTypeInfo> tryParseUuidExtensionMetadata(const ArrowSchema* schema,
+    const ArrowMetadataMap& metadata) {
+    // Arrow's UUID extension uses FixedSizeBinary(16) ("w:16") with an "ARROW:extension:name"
+    // of "arrow.uuid". See https://arrow.apache.org/docs/format/CDataInterface.html.
+    if (schema == nullptr || schema->format == nullptr) {
+        return std::nullopt;
+    }
+    if (schema->format[0] != 'w' || std::string(schema->format) != "w:16") {
+        return std::nullopt;
+    }
+    const auto extensionName = getMetadataValue(metadata, "arrow:extension:name");
+    if (!extensionName.has_value() || *extensionName != "arrow.uuid") {
+        return std::nullopt;
+    }
+    return ArrowLogicalTypeInfo{ArrowLogicalTypeInfo::Source::GENERIC_METADATA,
+        ArrowLogicalTypeInfo::Type::UUID, ArrowDecimalTypeInfo{0, 0}};
+}
+
 } // namespace common
 } // namespace lbug
--- a/src/common/arrow/arrow_type.cpp
+++ b/src/common/arrow/arrow_type.cpp
@@ -22,6 +22,8 @@ LogicalType ArrowConverter::fromArrowSchema(const ArrowSchema* schema) {
         case ArrowLogicalTypeInfo::Type::DECIMAL:
             return LogicalType::DECIMAL(logicalTypeInfo->decimal.precision,
                 logicalTypeInfo->decimal.scale);
+        case ArrowLogicalTypeInfo::Type::UUID:
+            return LogicalType::UUID();
         }
     }
     switch (arrowType[0]) {
--- a/src/include/common/arrow/arrow_schema_metadata.h
+++ b/src/include/common/arrow/arrow_schema_metadata.h
@@ -22,6 +22,7 @@ struct ArrowLogicalTypeInfo {
 
     enum class Type : uint8_t {
         DECIMAL,
+        UUID,
     };
 
     Source source;
--- a/src/storage/table/arrow_rel_table.cpp
+++ b/src/storage/table/arrow_rel_table.cpp
@@ -53,6 +53,38 @@ static std::vector<std::optional<ArrowLogicalTypeInfo>> resolveColumnLogicalType
     return result;
 }
 
+// Returns true when the Arrow endpoint column's wire type is acceptable for the
+// node table's primary key type. The historical rule is "exact match" via
+// toString(); a UUID primary key breaks that because no Arrow wire type round-trips
+// to a UUID logical type. The `arrow.uuid` extension on FixedSizeBinary(16) is
+// the canonical Arrow encoding of a UUID, so we accept that as well, with the
+// 16 bytes assumed to be in the big-endian (network) byte order that the
+// extension spec mandates. See https://arrow.apache.org/docs/format/CDataInterface.html.
+static bool isArrowEndpointCompatibleWithPK(const ArrowSchema* endpointSchema,
+    const LogicalType& pkType) {
+    if (endpointSchema == nullptr) {
+        return false;
+    }
+    auto arrowType = ArrowConverter::fromArrowSchema(endpointSchema);
+    if (arrowType.toString() == pkType.toString()) {
+        return true;
+    }
+    if (pkType.getLogicalTypeID() == LogicalTypeID::UUID) {
+        if (endpointSchema->format == nullptr) {
+            return false;
+        }
+        if (std::string(endpointSchema->format) != "w:16") {
+            return false;
+        }
+        if (auto logicalTypeInfo = tryGetArrowLogicalTypeInfo(endpointSchema);
+            logicalTypeInfo.has_value() &&
+            logicalTypeInfo->type == ArrowLogicalTypeInfo::Type::UUID) {
+            return true;
+        }
+    }
+    return false;
+}
+
 void ArrowRelTableScanState::setToTable(const transaction::Transaction* transaction, Table* table_,
     std::vector<column_id_t> columnIDs_, std::vector<ColumnPredicateSet> columnPredicateSets_,
     RelDataDirection direction_) {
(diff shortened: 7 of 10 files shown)


alvorithm:
Reporter of #757 here. Tested this branch at `62d10f224`, `relwithdebinfo`, C API only. The accepted path works end to end: the rel table loads, both endpoints resolve to the right nodes, and the payload column reads back intact.

Two things worth folding in.

**The reproducer attached to #757 cannot show this fix.** `arrow_rel_pk.c` builds its `FixedSizeBinary(16)` case with no field metadata, so against this branch all four of its cases still fail. That is correct behaviour, but anyone who re-runs the attached file to check the fix will read it as a failure. The new message makes the reason clear:

```
Runtime exception: Arrow 'from' column type w:16 is not compatible with source node PK
type UUID; supported: matching logical type, or FixedSizeBinary(16) with the
'arrow.uuid' extension for UUID
```

I wrote a second reproducer that builds the `ARROW:extension:name` metadata blob by hand and drives the accepted path. Happy to attach it.

**The UUID fixtures cannot catch a byte-order regression.** `ScanArrowRelTableOverNativeUuidNodeTable` uses `11111111-1111-1111-1111-111111111111`, `22222222-...` and `33333333-...`. Every byte inside each of those values is identical, so the 16-byte reversal in `scanArrowArrayUuid` is a no-op on them: the test passes whether or not the byte order is correct. The MSB flip is still covered, since flipping it changes the value. Only the ordering is invisible.

One asymmetric fixture closes the gap. I ran that against the branch:

| check | result |
|---|---|
| `w:16` with `arrow.uuid` | accepted |
| `w:16` without the extension (control) | rejected, as designed |
| edge lands on the right two nodes | `a -> b` |
| source id round trip | `01234567-89ab-cdef-0123-456789abcdef` |
| target id round trip | `fedcba98-7654-3210-fedc-ba9876543210` |

The byte order is right today. The suggestion is only about keeping it that way if the read path is ever touched again.

adsharma:
Thanks for the careful read.

**Byte-order fixture** — addressed in 213585e6f on the branch. Added `ScanArrowRelTableWithAsymmetricUuidEndpoints` alongside the symmetric-UUID test, using the same `01234567-89ab-cdef-0123-456789abcdef` / `fedcba98-7654-3210-fedc-ba9876543210` values you ran. The new test asserts both the edge resolves to the right nodes (a -> b) and that the UUIDs round-trip exactly via `MATCH ... RETURN a.id, b.id`, so any future byte-order regression in `scanArrowArrayUuid` will fail it.

**#757 reproducer** — agreed, that file can't show the fix because it builds the `FixedSizeBinary(16)` case without the `arrow.uuid` extension metadata. The new error message you quoted makes the reason clear, so anyone re-running it against this branch will see why every case still fails. The acceptance path is now covered by the two tests in the PR (`ScanArrowRelTableOverNativeUuidNodeTable` and the new asymmetric one) and by the three `ArrowConverterTest` schema-binding tests in `test/api/arrow_test.cpp`, all of which use the proper extension metadata. If you still want to attach the hand-rolled reproducer to #757, that would be a useful follow-up to clarify against the original report, but it doesn't need to block this PR.

PR #777, opened 2026-08-04
fix(build): make OpenSSL 3 optional — Windows builds fail without it

chiangchenghsin-hash:
## Summary

Changes `find_package(OpenSSL 3 REQUIRED)` to `QUIET` + `if (OpenSSL_FOUND)` guard, so that Windows builds and minimal CI environments can compile LadybugDB without OpenSSL 3 installed.

## Problem

v0.19.0 changed OpenSSL from optional to `REQUIRED`. This breaks:

1. **Windows MSVC builds** — OpenSSL 3 is not commonly installed on Windows (no `apt install libssl-dev` equivalent). The build fails with:
```
Could NOT find OpenSSL, try to set the path to OpenSSL root folder
(missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR)
```

2. **src/extension/CMakeLists.txt** — Even if the root CMakeLists.txt were patched locally, the extension CMake unconditionally references `${OPENSSL_INCLUDE_DIR}` which is `NOTFOUND`, causing a confusing secondary error.

### `CMakeLists.txt`

```cmake
# Before:
find_package(OpenSSL 3 REQUIRED)

# After:
find_package(OpenSSL 3 QUIET)
if (OpenSSL_FOUND)
    message(STATUS "OpenSSL 3 found — enabling HTTPS for extension downloads")
    ...
else()
    message(STATUS "OpenSSL 3 not found — building without HTTPS extension download support")
endif()
```

### `src/extension/CMakeLists.txt`

```cmake
# Before:
if (NOT BUILD_WASM)
    target_compile_definitions(lbug_extension PRIVATE CPPHTTPLIB_OPENSSL_SUPPORT)
    target_include_directories(lbug_extension PRIVATE ${OPENSSL_INCLUDE_DIR})
endif ()

# After:
if (NOT BUILD_WASM AND OpenSSL_FOUND)
    target_compile_definitions(lbug_extension PRIVATE CPPHTTPLIB_OPENSSL_SUPPORT)
    target_include_directories(lbug_extension PRIVATE ${OPENSSL_INCLUDE_DIR})
endif ()
```

## What still works without OpenSSL

- Core database: all CRUD operations, Cypher queries, graph algorithms
- Extensions: vector, algo, fts — all compile and run
- Extension loading from local paths: `LOAD EXTENSION './path/ext.lbug_extension'`

## What degrades without OpenSSL

- Extension download from remote repositories over HTTPS (uses httplib with SSL)
- The `extension_installer` falls back to HTTP or reports a clear error

## Verification

- Windows 11 x64 MSVC 19.50 without OpenSSL: full build (shell + vector + algo + fts extensions) succeeds
- Linux with OpenSSL 3: behavior unchanged (OpenSSL_FOUND = true, same as before)

## Note

This restores the v0.18.x behavior where OpenSSL was optional. The `REQUIRED` change in v0.19.0 appears to have been unintentional — the extension installer already handles the no-OpenSSL case gracefully.


chiangchenghsin-hash:
Adding some motivation for this change:

When building LadybugDB 0.19.0 from source on Windows 11 x64 (MSVC 19.50), we hit find_package(OpenSSL 3 REQUIRED) which blocks the build. Unlike Linux (e.g. apt install libssl-dev), Windows has no package manager to easily install OpenSSL 3, so the whole project fails to compile — including the core database and all extensions (vector, algo, fts).

Findings:

1. In v0.18.x OpenSSL was optional (QUIET + guard); v0.19.0 changed it to REQUIRED

2. OpenSSL is only needed for HTTPS extension downloads; the core database and all extensions compile and run without it extension.ladybugdb.com also supports HTTP fallback

3. With this fix we successfully compiled the full 0.19.0 release package on Windows (shell + vector + algo + fts extensions, 4 binaries, 56MB total), and verified ABI compatibility of our self-built extensions against the official prebuilt lbug.exe.


This change does not affect Linux/macOS environments that have OpenSSL (behavior is completely unchanged); it only enables builds on environments without OpenSSL.

adsharma:
Your commits are stripping all the comments for some reason. Can you please fix, so only the intended changes are included in the diff?

chiangchenghsin-hash:
Fixed. The branch was rebased on the current main; the diff now contains only the two intended OpenSSL changes (the `find_package(OpenSSL 3 REQUIRED)` → `QUIET` + guard in `CMakeLists.txt` and the `OpenSSL_FOUND` condition in `src/extension/CMakeLists.txt`). All unrelated formatting/comment changes have been reverted — sorry about the noise.

Verified on Windows 11 x64 / MSVC 19.50 without OpenSSL: full build (core + vector + algo + fts extensions) succeeds.

adsharma:
The code looks fine. But it's not mergeable due to git conflicts.

Suggest opening a new PR with identical content and a single commit.

chiangchenghsin-hash:
Thanks for the review and the clear guidance, @adsharma! Following your suggestion, I've reopened this as a **single commit** against the latest `main`:

- **New PR:** #796
- Same content (both files byte-identical to what you reviewed here)
- Single commit `245edbd9`, no merge commits, no conflicts

Appreciate the workflow tips from #773 as well — rebasing and keeping the diff focused from now on. 🙏

Issue #772, opened 2026-08-02
pokec medium: exact N hop reachability


adsharma:
Results from 0.19.0:

code: https://github.com/adsharma/falkordb-benchmark/tree/ladybug
data: https://huggingface.co/datasets/ladybugdb/pokec-medium/tree/main
output: https://gist.github.com/adsharma/bfb9b8828b55371a03a61feb2e42a936

Optimization opportunities:

* Avoid hash joins for count queries
* Look into better memory layout for faster traversal

adsharma:
Similar optimization: https://github.com/LadybugDB/ladybug/pull/646

PR #782, opened 2026-08-05
feat(alter): add SORTED BY ... CSR constraint 

adsharma:
Context: #646 

Replace the incorrect heuristic in count_rel_table_optimizer (which treated an ascending integer primary-key sort order as guaranteeing primary_key == rowid) with an explicit, user-declared CSR constraint.

ALTER TABLE t SORTED BY <col> (ASC) CSR;

The CSR keyword asserts that the node table's primary key is a csr_index that is interchangeable with the rel table's table_offset (primary_key == rowid). Because this is only a user declaration and not provable from the data, the optimizer now gates the offset-count degree rewrite (RelDegreeTableMode::OFFSET_COUNT) on:

  1. the CSR flag being set, and
  2. the node table not having been mutated (changeEpoch unchanged) since the CSR declaration. Any node-table mutation invalidates the invariant and the optimization is disregarded.

Validation: CSR requires exactly one SORTED BY property, ascending, and it must be the primary key. Composite/DESC/non-PK CSR declarations are rejected.

Plumbing:
- Grammar + ANTLR regeneration (new CSR keyword)
- Parser/binder/alter threading (csr + csrChangeEpoch)
- Node-table catalog metadata + storage version 44 persistence
- WAL serialization for the new fields
- Tests: optimizer CSR gating + mutation invalidation; e2e DDL validation


adsharma:
Also fixes: #772 

```
lbug> ALTER TABLE User set SORTED BY (id asc) CSR;
┌─────────────────────────────────────────────────────────┐
│ result                                                  │
│ STRING                                                  │
├─────────────────────────────────────────────────────────┤
│ Table User sorted-by metadata updated with 1 column(s). │
└─────────────────────────────────────────────────────────┘
(1 tuple)
(1 column)
Time: 0.53ms (compiling), 8.06ms (executing)
lbug> MATCH (a:User {id: 40692})-[Friend*1..5]->(b:User) RETURN count(distinct b);
┌───────────────────────┐
│ COUNT(DISTINCT b._ID) │
│ INT64                 │
├───────────────────────┤
│ 91457                 │
└───────────────────────┘
(1 tuple)
(1 column)
Time: 0.91ms (compiling), 131.91ms (executing)

Issue #16, opened 2025-10-17
CLI: What should the binary be called?


adsharma:
Something easy to type and not too contrived.

PR #785, opened 2026-08-06
c_api: exception floor (guard macro with catch-all) and first conversions

jonasvanderhaegen:
No function in `src/c_api/` currently has a `catch (...)`, and several files
have no exception handling at all, so `std::bad_alloc`, `std::system_error`,
or any exception from bundled third-party code escapes the C boundary and
terminates the embedding process. We embed lbug (skybox) and have field
evidence of aborts crossing this boundary.

This PR:
- adds `LBUG_C_API_GUARD_BEGIN` / `LBUG_C_API_GUARD_END(err_ret)` to
  `c_api/helpers.h`: `catch (const std::exception&)` (covers
  `lbug::common::Exception`, which derives from it) plus `catch (...)`,
  both routing through `setLastCAPIErrorMessage`;
- converts the fully-unguarded small files `query_summary.cpp` and
  `version.cpp`, adding null-argument checks where absent;
- null-checks the `malloc` in `convertToOwnedCString`.

Follow-ups (happy to do them as separate PRs to keep review small): apply the
macro across `value.cpp`, `query_result.cpp`, `prepared_statement.cpp`, and
extend the existing `catch (Exception&)` sites with the catch-all.

### DIFF
--- a/src/c_api/helpers.cpp
+++ b/src/c_api/helpers.cpp
@@ -71,6 +71,9 @@ char* takeLastCAPIErrorMessage() {
 char* convertToOwnedCString(const std::string& str) {
     size_t src_len = str.size();
     auto* c_str = (char*)malloc(sizeof(char) * (src_len + 1));
+    if (c_str == nullptr) {
+        return nullptr;
+    }
     memcpy(c_str, str.c_str(), src_len);
     c_str[src_len] = '\0';
     return c_str;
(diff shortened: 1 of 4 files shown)


adsharma:
Yes, this is much needed. Currently backed up on my review queue. Will get to this and your other PRs in a couple of days. Thank you for the contribution!

jonasvanderhaegen:
Closing: superseded by the full C API exception floor, which now lives in our fork (skylence-be/ladybug, merged as #16) rather than as an upstream contribution. Thanks.

Issue #692, opened 2026-07-15
REL TABLE GROUP property IN predicate duplicates rows and omits another endpoint table


n4s5ti:
## Summary

With `@ladybugdb/core` 0.16.1, a generic scan over a heterogeneous `REL TABLE GROUP` returns incorrect rows when a relationship property is filtered with a subset `IN` predicate.

For the same target and data:

- `r.type = 'CALLS'` returns the two expected, distinct rows.
- `r.type IN ['CALLS']` returns two copies of the `File -> Function` row and omits the `Function -> Function` row.
- `r.type IN ['CALLS', 'DEFINES']` returns all three expected rows.
- Adding explicit source labels (`s:File` or `s:Function`) returns the correct row.
- Binding the target id as `$target` instead of using a literal does not change the defect.
- `DISTINCT` hides the duplicate but cannot restore the omitted row.

An unfiltered generic scan returns all physical rows exactly once, so the index itself is not missing or duplicating these edges.

## Environment

- `@ladybugdb/core`: 0.16.1 (`@ladybugdb/core-linux-x64`)
- Node.js: 24.16.0
- OS: Arch Linux x86_64
- Index producer: GitNexus 1.6.5

## Reproduction

The executable harness is here:

https://github.com/n4s5ti/pip3r/blob/d41d2ac/packages/graphos/tests/ladybug-in-predicate-repro.ts

A fresh, forced index reproduces the issue, ruling out stale incremental state:

```bash
git clone https://github.com/n4s5ti/pip3r.git
cd pip3r
git checkout obs-1271-hotspot-hardening
npm ci
npx gitnexus analyze --force
cd packages/graphos
npx tsx tests/ladybug-in-predicate-repro.ts --production-index
```

Relevant output:

```json
{
  "equality": { "rows": 2, "distinct": 2 },
  "inSingle": { "rows": 2, "distinct": 1 },
  "inMulti": { "rows": 3, "distinct": 3 },
  "productionIndexBugReproduced": true
}
```

The incorrect `inSingle` rows are the same `File -> Function` `CALLS` edge twice; the `Function -> Function` `CALLS` edge is missing.

## Query matrix

Target:

```text
Function:packages/runtime/src/executor-wiring.ts:registerExecutor
```

Baseline / correct:

```cypher
MATCH (s)-[r:CodeRelation]->(t)
WHERE t.id = 'Function:packages/runtime/src/executor-wiring.ts:registerExecutor'
[...]

PR #786, opened 2026-08-06
ci: nightly sanitizer workflow (ASan+UBSan blocking, TSan signal-gathering)

jonasvanderhaegen:
The build system already plumbs `ASAN`/`TSAN`/`UBSAN` and `ENABLE_RUNTIME_CHECKS`
through CMake and the Makefile, but no workflow exercises them. This adds a
nightly (cron 06:00 UTC, off the 03:00 CI slot) with two linux legs:

- **asan-ubsan**: blocking. `RUNTIME_CHECKS=1 ASAN=1 UBSAN=1`, so the 1500+
  `DASSERT` invariants run under sanitizers nightly. Leak detection is off for
  the first landing to keep the signal actionable; it can be enabled once the
  baseline is clean.
- **tsan**: `continue-on-error`, pure signal-gathering. The engine is
  morsel-parallel and has never run under TSan in CI; this leg accumulates a
  baseline without blocking anyone.

Motivation from the consumer side: we ship lbug embedded (skybox) and
root-caused a silent wrong-results bug (#692) whose entire class fails
silently in release builds because `DASSERT` compiles out. A nightly with
runtime checks plus sanitizers is the cheapest structural counter.

### DIFF
--- a/.github/workflows/nightly-sanitizers.yml
+++ b/.github/workflows/nightly-sanitizers.yml
@@ -1,0 +1,47 @@
+name: Nightly Sanitizers
+
+on:
+  schedule:
+    - cron: '0 6 * * *'
+  workflow_dispatch:
+
+env:
+  PIP_BREAK_SYSTEM_PACKAGES: 1
+  RUNTIME_CHECKS: 1
+
+jobs:
+  asan-ubsan:
+    name: ASan + UBSan (linux, runtime checks on)
+    runs-on: ubuntu-24.04
+    timeout-minutes: 300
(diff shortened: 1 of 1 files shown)


adsharma:
@jonasvanderhaegen were you able to run the sanitizers in your env?

> Mmap for size 8796093022208 failed

This can be worked around with BM_MALLOC=1 because the 8TB virtual interacts badly with many of the sanitizers.

https://github.com/LadybugDB/ladybug/actions/runs/31422358105

PR #795, opened 2026-08-11
Add SIMD filter comparison kernels

rahul-iyer:
## Summary

Add specialized filter-selection kernels for fixed-width numeric comparisons.

The implementation accelerates dense, unfiltered filter inputs while preserving the existing generic executor as the fallback for unsupported shapes and types.

## What changed

- Add AVX2 kernels for x86-64.
- Add NEON kernels for ARM64.
- Support MSVC, GCC, and Clang builds.
- Add runtime AVX2 CPU detection on x86-64.
- Add `LBUG_ENABLE_SIMD`, enabled by default.
- Add `LBUG_SIMD` runtime selection:
  - `auto` / `1`: select the platform backend
  - `scalar` / `0`: force the scalar reference kernel
  - `avx2` or `neon`: force the corresponding platform backend
- Add dense fast paths to `BinaryFunctionExecutor` for:
  - vector-to-constant comparisons
  - constant-to-vector comparisons
  - vector-to-vector comparisons
- Support:
  - `INT8`–`INT64`
  - `UINT8`–`UINT64`
  - `FLOAT`
  - `DOUBLE`
  - equality and ordered comparisons
- Apply null masks to generated lane masks.
- Preserve existing floating-point/NaN comparison semantics.
- Fall back to the existing generic executor for filtered selection vectors,
  small inputs, unsupported types, and incompatible vector states.

## ARM64 optimization

The initial NEON implementation regressed at medium and high selectivity because
it stored comparison masks to the stack and compacted every vector separately.

This PR now:

- extracts lane masks directly from NEON registers;
- evaluates and compacts 32 rows at a time;
- special-cases empty and all-selected masks;
- writes contiguous selection positions with vector stores;
- uses branchless dense scalar compaction for 64-bit vector-to-vector comparisons,
  where two-lane NEON was slower.

## Performance

Apple M5 Pro, Release build, one thread, warm 10-million-row dataset.

Each result is the median of 100 interleaved SIMD-on/off executions. Query
compilation and dataset creation are excluded.

| Filter | Matches | SIMD off | Optimized | Improvement |
|---|---:|---:|---:|---:|
| `value < 10` | 1% | 5.851 ms | 4.964 ms | 17.9% |
| `value < 100` | 10% | 5.806 ms | 4.981 ms | 16.6% |
| `value < 500` | 50% | 5.824 ms | 5.089 ms | 14.4% |
| `value < 900` | 90% | 5.782 ms | 5.290 ms | 9.3% |
| `value < other` | ~50% | 7.473 ms | 7.219 ms | 3.5% |

All benchmark result counts matched the expected values.

The final ARM64-enabled benchmark executable is approximately 166 KB larger
than the otherwise-identical SIMD-disabled Release executable.

AVX2 performance has not yet been measured on x86 hardware.

## Testing

Added differential tests covering:

- every supported integer width;
- signed and unsigned comparisons;
- `FLOAT` and `DOUBLE`;
- all six comparison operations;
- null masks and null-word boundaries;
- floating-point infinities, signed zero, and NaNs;
- vector-to-constant comparisons in both operand directions;
- dense vector-to-vector comparisons.

Verified locally:

```text
LBUG_SIMD=neon types_test --gtest_filter='SIMDFilterTest.*'
4 tests passed

LBUG_SIMD=scalar types_test --gtest_filter='SIMDFilterTest.*'
4 tests passed

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,7 @@ option(ENABLE_UBSAN "Enable undefined behavior sanitizer." FALSE)
 option(ENABLE_RUNTIME_CHECKS "Enable runtime coherency checks (e.g. asserts)" FALSE)
 option(ENABLE_LTO "Enable Link-Time Optimization" FALSE)
 option(ENABLE_MALLOC_BUFFER_MANAGER "Enable Buffer manager using malloc. Default option for webassembly" OFF)
+option(LBUG_ENABLE_SIMD "Enable runtime-dispatched query-processor SIMD kernels." ON)
 
 option(LBUG_DEFAULT_REL_STORAGE_DIRECTION "Only store fwd direction in rel tables by default." BOTH)
 if(NOT LBUG_DEFAULT_REL_STORAGE_DIRECTION)
--- a/scripts/headers.txt
+++ b/scripts/headers.txt
@@ -47,6 +47,8 @@ src/include/common/vector/auxiliary_buffer.h
 src/include/common/vector/value_vector.h
 src/include/function/binary_function_executor.h
 src/include/function/cast/cast_function_bind_data.h
+src/include/function/comparison/comparison_operation.h
+src/include/function/comparison/simd_filter.h
 src/include/function/const_function_executor.h
 src/include/function/function.h
 src/include/function/pointer_function_executor.h
--- a/src/function/CMakeLists.txt
+++ b/src/function/CMakeLists.txt
@@ -19,6 +19,43 @@ add_subdirectory(export)
 add_subdirectory(internal_id)
 add_subdirectory(timestamp)
 
+set(LBUG_FUNCTION_SIMD_SOURCES simd_filter.cpp)
+set(LBUG_BUILD_FILTER_AVX2 FALSE)
+set(LBUG_BUILD_FILTER_NEON FALSE)
+if(LBUG_ENABLE_SIMD AND NOT BUILD_WASM)
+    if(OS_ARCH STREQUAL "amd64")
+        if(MSVC)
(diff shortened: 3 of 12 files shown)


adsharma:
Sorry about the GQL CI breakage. Rebasing should make it go away.

adsharma:
One of the criticisms I've heard about duckdb is the 70MB binary size. We're at around 10-12MB range last I looked. Would be good to quantify the hit to figure out where the sweet spot is in terms of cost/benefit.

adsharma:
## Text size comparison (before vs after)

Measured with `readelf -SW` on the built artifacts.

### Executable `lbug`

| Section | before | after | delta |
|---|---|---|---|
| .text | 12,995,682 | 13,158,434 | **+162,752 (+1.25%)** |
| .rodata | 1,140,936 | 1,143,112 | +2,176 (+0.19%) |
| .bss | 21,232 | 21,264 | +32 |

### Static archive `liblbug.a` (sum of member `.text`)

7,050,012 → 7,197,288 = **+147,276 (+2.09%)**

The entire archive delta comes from two new objects — no existing member changed size, none were removed:

| object | .text |
|---|---|
| `simd_filter_avx2.cpp.o` | +110,962 |
| `simd_filter.cpp.o` | +36,314 |
| **total** | **+147,276** |

### Notes

- The linked binary's +162,752 is ~15.5 KB larger than the archive sum, which is expected from link-time codegen/inlining of the new SIMD kernels, plus the +2,176 rodata (likely constants/lookup tables for the SIMD paths) and +32 bss.
- ~111 KB of .text in `simd_filter_avx2.cpp.o` is typical for an AVX2 kernel file (unrolled loops, multiple filter variants).

PR #807, opened 2026-08-12
fix(parser): collapse a doubled backtick in an escaped identifier

### LINKED ISSUE #770: Bug: doubled backticks in an escaped identifier are not collapsed
An escaped identifier that contains a doubled backtick keeps both backticks instead of collapsing them to one. Verified against a `relwithdebinfo` build of `4ca1d6f5d`. This was noted inside #756 as the reason its patch does not escape; it now has a reproducer, so here it is separately.

## What happens

```
CREATE NODE TABLE `a``b`(id INT64, PRIMARY KEY(id));
CALL SHOW_TABLES() RETURN *;
```

```
┌────────┬────────┬────────┬───────────────┬─────────┐
│ id     │ name   │ type   │ database name │ comment │
├────────┼────────┼────────┼───────────────┼─────────┤
│ 0      │ a``b   │ NODE   │ main(graph)   │         │
└────────┴────────┴────────┴───────────────┴─────────┘
```

The table is named `a``b`, with two backticks. Under the usual doubling convention the name asked for is `a`b`, with one.

The result is self-consistent: `CREATE (:`a``b` {id: 1})` and `MATCH (n:`a``b`) RETURN count(n)` both work and return 1. So nothing is broken for a user who never round-trips a name through generated text.

## Where

`Transformer::transformSymbolicName` (`src/parser/transformer.cpp:212`) strips the outer pair and nothing else:

```cpp
std::string escapedSymbolName = ctx.EscapedSymbolicName()->getText();
// escapedSymbolName symbol will be of form "`Some.Value`". Therefore, we need to sanitize
// it such that we don't store the symbol with escape character.
return escapedSymbolName.substr(1, escapedSymbolName.size() - 2);
```

The grammar is already built for the doubled form. `src/antlr4/Cypher.g4:855`:

```
EscapedSymbolicName
    : ( '`' ( EscapedSymbolicName_0 )* '`' )+ ;
fragment EscapedSymbolicName_0 : ~[`] ;
```

The trailing `+` is what lets ``` `a``b` ``` lex as a single token, as two adjacent quoted groups. That repetition has no other purpose, and `Cypher.g4:769` already special-cases the empty ``` `` ``` token by name. So the grammar treats doubling as the escape and the transformer does not.

`git log -L 212,222:src/parser/transformer.cpp v0.19.0..HEAD` is empty.

## Why it is worth fixing

Any code that quotes an identifier by wrapping it in backticks emits unparseable Cypher when the name already contains one. The DDL generator added in #761 does exactly that, so an Arrow field named ``` `index` ``` (backticks part of the name) now fails:

```
Parser exception: mismatched input '``' expecting PRIMARY (line: 1, offset: 38)
"CREATE NODE TABLE `t15` (`id` STRING, ``index`` STRING, PRIMARY KEY (`id`)) WITH (storage='arrow://arrow_15')"
                                       ^^
```

Every other name in that sweep passes after #761, including `index`, `order`, `my column`, `my-column` and `2nd`. This is the one remaining shape, and it is reachable from any Arrow producer that lets a user name a column.

## Fix options, your call

1. **Collapse the doubling in `transformSymbolicName`.** Strip the outer pair, then replace each ``` `` ``` with a single backtick. This makes the grammar and the transformer agree, and makes a backtick expressible in an identifier for the first time. It changes the meaning of existing queries that use the doubled form, so it is a breaking change for anyone relying on today's behaviour.
2. **Reject the doubled form** with a clear message rather than accepting it and mangling it. Cheaper, and it makes the limitation explicit instead of silent, but it leaves backticks unrepresentable.
3. **Leave the parser alone and make the writers escape.** Wherever an identifier is interpolated into generated DDL, double any backtick in it. That fixes the #761 symptom without touching query semantics, but the round trip still does not hold.

Option 1 with option 3 is the only combination that makes a generated name survive a round trip.

## Tests

`test/test_files/ddl/ddl.test` is where a name case fits. Whichever option you take, the case worth pinning is: create a table whose name contains a backtick, read the name back through `SHOW_TABLES`, quote it again and re-parse it.

## Notes

`lbug` built from `4ca1d6f5d`, `make relwithdebinfo NUM_THREADS=14`. Shell reproducer above is plain Cypher, no bindings involved. The Arrow failure comes from `arrow_ddl.c`, already attached to #756, run against the same build.

alvorithm:
_This work was produced with the help of language models._

Fixes #770.

The lexer reads a whole escaped name, doublings included, as one token:

    EscapedSymbolicName : ( '`' ( ~[`] )* '`' )+ ;

`Transformer::transformSymbolicName` strips the outer pair and stores what is left, which keeps the doubling. The name a writer asks for and the name the catalogue holds differ:

    written in Cypher   `a``b`
    stored on main      a``b
    stored here         a`b

The stored form on `main` is self-consistent for a caller who only types the name. It breaks the moment a name makes the round trip, because a writer that quotes by wrapping in backticks then emits Cypher the lexer reads as a different name. That is how the Arrow DDL generator fails on a field whose name contains a backtick, which is where this was found.

The issue offers three options and asks for a decision. This takes the first with the third, because that pair is the only combination under which a generated name survives the trip.

## The parser collapses the doubling

`transformSymbolicName` walks the inner text and emits one backtick for each doubled pair, which is the escape the grammar's repetition encodes. Every other identifier form keeps its old path.

## Every writer of generated Cypher doubles what it writes

`common::StringUtils::quoteIdentifier` wraps a name and doubles any backtick inside it. Seventeen call sites that interpolate a catalogue name into Cypher now go through it:

| file | what it writes |
|---|---|
| `catalog/catalog_entry/node_table_catalog_entry.cpp` | `CREATE NODE TABLE`, and its primary key name |
| `catalog/catalog_entry/rel_group_catalog_entry.cpp` | `CREATE REL TABLE`, and its `FROM` and `TO` names |
| `catalog/catalog_entry/index_catalog_entry.cpp` | `CREATE INDEX`, its table and its property |
| `catalog/catalog_entry/sequence_catalog_entry.cpp` | `CREATE SEQUENCE` and `DROP SEQUENCE` |
| `catalog/catalog_entry/graph_catalog_entry.cpp` | `CREATE GRAPH` and `DROP GRAPH` |
| `catalog/property_definition_collection.cpp` | the property list inside those statements |
| `binder/bind/bind_export_database.cpp` | the `MATCH` that reads each table out |
| `processor/operator/simple/export_db.cpp` | the `COPY` statements in the exported script |
| `binder/bind/bind_import_database.cpp` | the `COPY` statements it re-binds on import |
| `function/scalar_macro_function.cpp` | `CREATE MACRO` |
| `function/gds/gds.cpp` | the projection queries behind the graph functions |
| `function/table/project_native_graph.cpp` | the projection query for a native graph |
| `storage/table/arrow_table_support.cpp` | the DDL for an Arrow-backed table |
| `test/test_runner/fsm_leak_checker.cpp` | the `DROP TABLE` the harness runs between cases |

The last one is a test file, and it is in the list because it failed first: with the parser fixed and the writer naive, the harness could no longer drop the table the new test creates.

The one place that keeps its bare backticks is the type-mismatch message in `reader_bind_utils.cpp`, where they are decoration in prose rather than an identifier the parser will read.

## The behaviour change

The catalogue holds a different name after this patch, for names that contain a backtick:

    CREATE NODE TABLE `a``b` (...)   ->  a`b   here
                                     ->  a``b  on main

A database written by an older build therefore reports one name and a new build reports another. Names without a backtick are unaffected.

## Tests

Both cases are in `test/test_files/ddl/ddl_empty.test`, and both fail on `main` at `41f704d05`.

`EscapedIdentifierCollapsesDoubledBacktick` creates a node table and a property whose names contain a backtick, reads both back through `SHOW_TABLES` and `TABLE_INFO`, then quotes them again in a `CREATE` and a `MATCH`. On `main` the first row reads:

    Table a``b has been created.

`EscapedIdentifierSurvivesExportImport` runs the same schema through `EXPORT DATABASE` and `IMPORT DATABASE` and reads the name and the property back. On this branch before the writers were converted, the export failed inside its own generated query:

    Parser exception: Invalid input <match (a:`x`y>: expected rule oC_SingleQuery (line: 1, offset: 12)

which is the round trip this pair of options is for.

`make test NUM_THREADS=14 TEST_JOBS=8`: 2672 passed and 4 failed out of 2676, on the rebase onto `41f704d05`. The four are `dictionary_bug~orb383_relationship_projection_obfuscated.AnonymousParquetDeleteReload`, which needs a gitignored Parquet fixture, and three `extension` cases that download from `extension.ladybugdb.com`. The same four fail on `main` on this host.

### DIFF
--- a/src/storage/table/arrow_table_support.cpp
+++ b/src/storage/table/arrow_table_support.cpp
@@ -5,6 +5,7 @@
 
 #include "common/arrow/arrow_converter.h"
 #include "common/exception/runtime.h"
+#include "common/string_utils.h"
 #include "main/database.h"
 
 namespace lbug {
@@ -46,7 +47,7 @@ static int64_t findArrowColumnByName(const ArrowSchemaWrapper& schema, const std
 // names are arbitrary strings, and many ordinary ones (`index`, `order`, `group`, names with a
 // space or a leading digit) are not valid unquoted Cypher identifiers.
 std::string quoteIdent(const std::string& name) {
-    return "`" + name + "`";
+    return common::StringUtils::quoteIdentifier(name);
 }
 
 std::string ArrowTableSupport::registerArrowData(ArrowSchemaWrapper schema,
--- a/src/binder/bind/bind_export_database.cpp
+++ b/src/binder/bind/bind_export_database.cpp
@@ -58,14 +58,17 @@ void bindExportTableData(ExportedTableData& tableData, const std::string& query,
 }
 
 static std::string getExportNodeTableDataQuery(const TableCatalogEntry& entry) {
-    return std::format("match (a:`{}`) return a.*", entry.getName());
+    return std::format("match (a:{}) return a.*", StringUtils::quoteIdentifier(entry.getName()));
 }
 
 static std::string getExportRelTableDataQuery(const TableCatalogEntry& relGroupEntry,
     const NodeTableCatalogEntry& srcEntry, const NodeTableCatalogEntry& dstEntry) {
-    return std::format("match (a:`{}`)-[r:`{}`]->(b:`{}`) return a.{},b.{},r.*;",
-        srcEntry.getName(), relGroupEntry.getName(), dstEntry.getName(),
-        srcEntry.getPrimaryKeyName(), dstEntry.getPrimaryKeyName());
+    return std::format("match (a:{})-[r:{}]->(b:{}) return a.{},b.{},r.*;",
+        StringUtils::quoteIdentifier(srcEntry.getName()),
+        StringUtils::quoteIdentifier(relGroupEntry.getName()),
+        StringUtils::quoteIdentifier(dstEntry.getName()),
+        StringUtils::quoteIdentifier(srcEntry.getPrimaryKeyName()),
+        StringUtils::quoteIdentifier(dstEntry.getPrimaryKeyName()));
 }
 
 static std::vector<ExportedTableData> getExportInfo(const Catalog& catalog,
--- a/src/binder/bind/bind_import_database.cpp
+++ b/src/binder/bind/bind_import_database.cpp
@@ -3,6 +3,7 @@
 #include "common/copier_config/csv_reader_config.h"
 #include "common/exception/binder.h"
 #include "common/file_system/virtual_file_system.h"
+#include "common/string_utils.h"
 #include "main/client_context.h"
 #include "parser/copy.h"
 #include "parser/parser.h"
@@ -43,7 +44,7 @@ static std::string getColumnNamesToCopy(const CopyFrom& copyFrom) {
     std::string delimiter = "";
     for (auto& column : copyFrom.getCopyColumnInfo().columnNames) {
         columns += delimiter;
-        columns += "`" + column + "`";
+        columns += common::StringUtils::quoteIdentifier(column);
         if (delimiter == "") {
             delimiter = ",";
         }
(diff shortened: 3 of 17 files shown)


src/storage/table/arrow_table_support.cpp
```diff
@@ -46,7 +47,7 @@ static int64_t findArrowColumnByName(const ArrowSchemaWrapper& schema, const std
 // names are arbitrary strings, and many ordinary ones (`index`, `order`, `group`, names with a
 // space or a leading digit) are not valid unquoted Cypher identifiers.
 std::string quoteIdent(const std::string& name) {
```
adsharma, line 49: future cleanup: remove this and inline?

PR #828, opened 2026-08-21
Optimize skewed intersect lists

rahul-iyer:
Ladybug currently intersects two sorted adjacency lists using a scalar merge. When one list is much smaller than the other, this can scan most of the larger list one element at a time.

This PR adds a galloping-search fast path for skewed intersections. It jumps through the larger list exponentially and then uses binary search to find each value from the smaller list.

Galloping is selected automatically when:

- The larger list has at least 64 entries.
- The larger list is at least 8× the size of the smaller list.
- Both lists contain nodes from the same table.

Small, balanced, and mixed-table lists continue using the original scalar merge.

Isolated kernel performance

These numbers measure only list copying and intersection, not complete query execution.

| List sizes | Kernel speedup across overlap levels |
|---|---:|
| 16 × 128 | 2.34×–2.70× |
| 16 × 256 | 2.99×–3.96× |
| 32 × 2048 | 7.45×–7.86× |
| 256 × 2048 | 2.24×–2.49×


adsharma:
Looks good. Some alternative designs if someone were to revisit:

**Planner-level fix ("pick HJ under skew")** — more general in principle, but skew here is per-chunk runtime state (which probe tuple meets which 2048-chunk), invisible to plan-time cardinality estimates. The existing TODO in cost_model.cpp shows this is known-hard. Not actionable now.

**Runtime-adaptive escalation** (start scalar; switch to galloping after N consecutive right-advances) — arguably more maintainable.

**SIMD/shuffle techniques** — premature at 2048-element scale; nodeID_t is 16 bytes so it would also require a repack after the homogeneity check.

adsharma:
Minor:

**LBUG_API** on both functions exports processor internals into the shared ABI purely so a test can reach them. Since the kernels are tiny and called from one hot site, consider a header-only implementation in src/include (inlining also helps here — the call is inside the vector-at-a-time hot loop), which removes the export and the new .cpp entirely. Minor, but it simplifies the file layout the PR adds.

Also could you run a cypher query that shows the benefit vs micro benchmark?

adsharma (changes requested):
Requesting changes for the `LBUG_API` and potential refactoring into a header only.

adsharma (approved):

adsharma:
Confirmed this is Baeza-Yates-style galloping (exponential probe + binary search in the bracket). I checked CRoaring reuse and measured an alternative — details below.

## Can we use something from CRoaring?

CRoaring is already vendored (`third_party/roaring_bitmap`) and its `array_container_intersection` has the same two-regime shape (skewed path at 64x ratio, SIMD merge otherwise). But three mismatches block direct reuse:

1. **Output contract.** Our kernel compacts matches into `left[]` *and* records both sides' selection positions for payload lookup. CRoaring only emits values — re-deriving positions costs a second pass and erases the win.
2. **Element layout.** CRoaring works on packed `uint16_t` arrays; we gallop on `offset_t = uint64_t` strided inside 16-byte `nodeID_t`s. Any adoption is a port, not a call.
3. **Not even the same algorithm.** CRoaring's skewed path (`intersect_skewed_uint16`) is *not* galloping — it's batched branchless binary search (`binarySearch4`/`2`) over the whole remaining suffix, with a comment noting it "can be slightly superior to a conventional galloping search in some instances."

## Third arm + benchmark (pushed to this branch)

I added `intersectSameTableBatchedSearch` to `intersect_kernels.h` — one branchless binary search per left element over the monotonically-shrinking right suffix (the honest port of CRoaring's approach to our layout/contract), behind the same dispatcher gating, plus differential-test coverage and a `BenchmarkSkewedPaths` test. ns per intersect, R=2048, release build, 2000 iters x 3 seeds:

| L | density | matches | scalar | gallop | batched | gallop/scalar | batched/scalar |
|---|---------|---------|--------|--------|---------|---------------|----------------|
| 8 | sparse | 0 | 9651 | 110 | 56 | 87x | 172x |
| 32 | sparse | 0 | 10614 | 349 | 218 | 30x | 49x |
| 256 | sparse | 3 | 12676 | 1457 | 1634 | 8.7x | 7.8x |
| 8 | dense | 3 | 9790 | 118 | 57 | 83x | 172x |
| 32 | dense | 13 | 12836 | 490 | 311 | 26x | 41x |
| 256 | dense | 100 | 12071 | 1376 | 1645 | 8.8x | 7.3x |

(Sparse: domain = 64x R, almost no matches — pure skip regime. Dense: domain = 2x R. Numbers stable across runs.)

## Takeaway

- Both fast paths crush scalar (8–170x) wherever the skew gate fires, so the optimization itself is clearly worthwhile.
- **Batched beats galloping ~1.6x at tiny L (8–32)** — exponential probing overhead dominates when the left list fits in a couple of cache lines. **Galloping retakes ~1.15x at L=256**, where the bracketed search pays off over repeated full-suffix binary searches.
- CRoaring's remark holds up on our layout: binary search *is* superior to galloping in the extreme-skew corner.

## Suggestion

Keep galloping as the default (it wins the mid-skew regime and matches the PR's intent), and consider a size-based pick in the dispatcher as a follow-up — e.g. batched search when `leftCount <= 32`, galloping above. That's a ~5-line change; happy to add it here if you want it in this PR, otherwise the third arm stays as differential-test coverage + the benchmark documents the tradeoff.

Issue #831, opened 2026-08-22
Feature: support single-argument properties() function for Node, Rel, and Struct


Mrhs121:
### API

C++

### Summary
Add support for the openCypher / ISO GQL standard single-argument `properties()` function on `NODE`, `REL`, and `STRUCT` expressions (e.g., `PROPERTIES(n)` and `PROPERTIES(r)`).

---

### Background & Motivation
Currently, Ladybug only supports the two-argument form of the `PROPERTIES()` function:
```cypher
PROPERTIES(nodes(p), 'propName')
```
While this is useful for variable-length path property extraction, openCypher and other Cypher implementations (such as [Neo4j](https://neo4j.com/docs/cypher-manual/current/functions/scalar/#functions-properties) and so on) standardly support a single-argument `properties(entity)`:
```cypher
MATCH (n:Person {name: 'Alice'})
RETURN properties(n);
```
Previously in Ladybug, calling `properties(n)` resulted in a binder error:
```text
Binder exception: Function PROPERTIES did not receive correct arguments:
Actual:   (NODE)
Expected: (LIST,STRING) -> ANY
```

---

### Key Use Cases
1. **Clean Payloads for APIs / JSON Serialization**: `RETURN n` returns graph entities containing internal system metadata (`_ID`, `_LABEL`). `properties(n)` extracts only clean user-defined business properties as a `STRUCT`.
2. **First-class Expression Composition**: Unlike `n.*` (which is a statement-level wildcard expansion and cannot be nested), `properties(n)` is a first-class expression that can be passed to aggregation functions like `collect(properties(n))`.
3. **GraphRAG & LLM Prompt Construction**: In Knowledge Graph / RAG pipelines, extracting clean entity/relationship property maps without system noise significantly reduces prompt token usage.

---

### Proposed Behavior
- `PROPERTIES(node: NODE) -> STRUCT`: Returns a struct containing all user-defined properties of the node (excluding internal columns `_ID`, `_LABEL`).
- `PROPERTIES(rel: REL) -> STRUCT`: Returns a struct containing all user-defined properties of the relationship (excluding internal columns `_ID`, `_LABEL`, `_SRC`, `_DST`).
- `PROPERTIES(struct: STRUCT) -> STRUCT`: Returns the struct with user fields.
- `PROPERTIES(list: LIST, prop: STRING) -> LIST`: Existing two-argument form remains fully backwards compatible.

---

### Example
```cypher
[...]

PR #832, opened 2026-08-22
feat: support single-argument properties() for node, rel, and struct

Mrhs121:
related to https://github.com/LadybugDB/ladybug/issues/831

### DIFF
--- a/src/function/path/properties_function.cpp
+++ b/src/function/path/properties_function.cpp
@@ -89,11 +89,17 @@ static void singleArgCompileFunc(FunctionBindData* bindData,
     const std::vector<std::shared_ptr<ValueVector>>& parameters,
     std::shared_ptr<ValueVector>& result) {
     DASSERT(parameters[0]->dataType.getPhysicalType() == PhysicalTypeID::STRUCT);
-    auto& propertiesBindData = bindData->cast<PropertiesBindData>();
-    for (auto i = 0u; i < propertiesBindData.fieldIndices.size(); ++i) {
-        auto inFieldVector =
-            StructVector::getFieldVector(parameters[0].get(), propertiesBindData.fieldIndices[i]);
-        if (inFieldVector->state == result->state) {
+    // Contract:
+    // If parameters[0] and result share the same DataChunkState (the common unflat pipeline case),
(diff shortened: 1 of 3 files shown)


adsharma:
Nice feature overall. Two correctness concerns and a nit:

1. **Null propagation:** neither `singleArgCompileFunc` nor `singleArgExecFunc` writes the result's outer null mask. For an unflat input (the common case) exec hits the sel-vector equality check and skips entirely, so `OPTIONAL MATCH (p:person {ID: 999}) RETURN properties(p)` yields a **non-null** struct — and on the reference path its fields point at the null node's uninitialized child vectors (possibly stale data). openCypher says `properties(null)` is `null`. Please propagate outer nulls (or always do a sel-aligned copy) and add an OPTIONAL MATCH regression test.

2. **Fragile invariant:** correctness relies on `resolveResultStateFromChildren` assigning the child's state *before* `compileFunc` runs, so the `state ==` check in compile matches the sel-pointer check in exec; if they ever disagree, the result is silently garbage rather than an error. Please document the contract, or unify both checks on `parameters[0]->state.get() == result.state.get()`.

Nit: `if (parameterSelVector == resultSelVector) continue;` in `singleArgExecFunc` is loop-invariant — hoist it above the loop.

adsharma (changes requested):
Looks good. Requesting changes for the `OPTIONAL MATCH` and `null` handling. The other two comments are minor.

Mrhs121:
Thanks for your review! I have addressed all the comments in the latest commit:

1. **Null propagation**: Added outer null mask propagation in `singleArgExecFunc` using `setNullFromBits` for unflat vectors and `setNull` for flat/filtered vectors, ensuring `properties(null)` evaluates to `null`. Added regression tests for `OPTIONAL MATCH` (null node / null relationship) and `properties(null)`.
2. **Contract & state check**: Documented the zero-copy referencing contract in `singleArgCompileFunc` and unified the state check on `parameters[0]->state.get() == result->state.get()`.
3. **Loop-invariant hoisted**: Hoisted the `parameterSelVector == resultSelVector` check out of the loop as an early return for the reference path.

PR #833, opened 2026-08-23
docs: fix broken threat model link in SECURITY.md and a typo

vaibhav8a:
Two small docs fixes, verified against the current tree on `main`.

**Broken link.** `SECURITY.md:17` points at `security/threat-model.md`, but the file in the repository is `security/threat_model.md` — hyphen vs. underscore, so the link 404s.

**Typo.** `tools/shell/shell_development_guide.md:138` — "due to the whole query being avaiable" → "available".

---

One thing I did **not** change, since the fix isn't obvious: `docs/testing.md:42` links to `../tools/nodejs_api/docs/nodejs_testing.md` as the "Node.js API — Testing Guide". `tools/nodejs_api/` exists, but it has no `docs/` directory and no markdown files at all, so there's no obvious intended target. Let me know if the guide moved or is still to be written and I'll follow up.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -14,4 +14,4 @@ To report a security issue, please email [security@ladybugdb.com](mailto:securit
 
 ## Threat Model
 
-Ladybug's current threat model is documented in [security/threat-model.md](security/threat-model.md).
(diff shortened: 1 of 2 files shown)


adsharma:
Thank you for the fixes!

PR #834, opened 2026-08-23
Partitioning: refuse RANGE until implemented; add dynamic LIST partitioning

adsharma:
## Summary

Two changes to table partitioning:

### 1. RANGE is refused at DDL time
RANGE previously fell back to hash routing while accepting `PARTITION BY RANGE`, silently violating the monotonic placement users expect. Real range partitioning needs bounds derived from the actual value distribution (declarative bounds or data-driven splitting); equal splits of the type domain were rejected because all realistic data lands in one middle bucket. The binder now refuses the DDL with an actionable message; the grammar keeps parsing it so enabling it later is a binder-only change.

### 2. LIST partitioning (one partition per distinct value)
`PARTITION BY LIST (col)` creates one partition per distinct partition-key value, dynamically:

- **Grammar**: new `iC_PartitionList` rule + `LIST` keyword (no PARTITIONS clause); regenerated parser sources and shell keywords.
- **Catalog**: LIST parents persist an encoded-key -> child-table-ID map on the parent entry (**storage version 47**; 0.20.0 repointed since v46 never shipped).
- **Dynamic creation**: a `ListPartitionRouter` creates a partition at first sight of a new value *inside the writing transaction*, via the same machinery as CREATE NODE TABLE (catalog entry + serial sequence + subgraph + storage + WAL create record), so rollback, WAL replay, and checkpointing behave like ordinary DDL. Children carry their parent link through the create info so replay reconstructs them correctly.
- **Routing**: single-row INSERT/MERGE resolve lazily in NodeInsertExecutor; COPY resolves under the router lock and atomically grows shared + per-worker target arrays (including index producer tokens) as workers discover values.
- One initial unkeyed partition is created at DDL so reads/writes always see a non-empty child set; it stays empty by design.

HASH routing is unchanged; all previously merged invariants hold (no direct child drops/alters, cascade-drop guards, partition-column updates refused, rel FROM/TO expands over partitions, parent rename cascades).

## Testing

- New e2e case `PartitionedListDynamic`: per-value partition creation on inserts, parent-scan union, partition reuse for repeated values, persistence across reopen/WAL replay, post-reopen routing into an existing partition, cascade drop.
- All 12 partitioned + storage_version e2e tests pass.
- Verified manually in the shell across close/reopen cycles.


adsharma:
Continued in #838

PR #872, opened 2026-08-31
fix iceberg extension REST catalog file binder exception

### LINKED ISSUE #871: Bug: iceberg extension REST catalog file Binder exception
### Ladybug version

0.20.0

ubuntu24

### What happened?

lbug> install iceberg
;
┌───────────────────────────────────────────────────────────────────────────────┐
│ result                                                                        │
│ STRING                                                                        │
├───────────────────────────────────────────────────────────────────────────────┤
│ Extension: iceberg installed from the repo: [https://extension.ladybugdb.com/](https://extension.ladybugdb.com/). │
└───────────────────────────────────────────────────────────────────────────────┘
(1 tuple)
(1 column)
Time: 1.38ms (compiling), 147820.22ms (executing)
lbug> load iceberg;
┌─────────────────────────────────────┐
│ result                              │
│ STRING                              │
├─────────────────────────────────────┤
│ Extension: iceberg has been loaded. │
└─────────────────────────────────────┘
(1 tuple)
(1 column)
Time: 0.27ms (compiling), 13.37ms (executing)
lbug> CALL iceberg\_warehouse='warehouse';
(0 tuples)
(0 columns)
Time: 3.05ms (compiling), 0.28ms (executing)
lbug> CALL iceberg\_endpoint='[http://127.0.0.1:8181](http://127.0.0.1:8181)';
(0 tuples)
(0 columns)
Time: 0.08ms (compiling), 0.34ms (executing)
lbug> CALL iceberg\_authorization\_type='none';
(0 tuples)
(0 columns)
Time: 0.12ms (compiling), 0.43ms (executing)
lbug> LOAD FROM 'perf\_demo.User'
..> (file\_format='iceberg')
..> RETURN count(\*);
Error: Binder exception: No file found that matches the pattern: perf\_demo.User.

### Problem

The Iceberg extension supports querying tables through an Iceberg REST Catalog using names such as:

```text
namespace.table
iceberg_catalog.namespace.table
```

However, `LOAD FROM` treats every source as a filesystem path during the core binding phase. Before the Iceberg extension receives the source, `bindFileScanSource()` calls `bindFilePaths()`, which performs local filesystem globbing.

For a REST Catalog table such as:

```cypher
LOAD FROM 'perf_demo.User' (file_format='iceberg') RETURN count(*);
```

the binder attempts to find a local file named:

```text
perf_demo.User
```

Because this file does not exist, binding fails with:

```text
No file found that matches the pattern: perf_demo.User.
```

The Iceberg extension REST Catalog resolver is never invoked.

_No response_

ericyuanhui:
fix https://github.com/LadybugDB/ladybug/issues/871  

### Root Cause

The REST Catalog implementation was added in the Iceberg extension, but the core `LOAD FROM` binder still assumes that the source must be a filesystem path.

The failure occurs before the extension table function is bound:

```text
LOAD FROM
  -> bindFileScanSource()
  -> bindFilePaths()
  -> filesystem glob
  -> binder error
```

### Solution

Update `bindFileScanSource()` to recognize qualified Iceberg REST Catalog table names before filesystem globbing.

When the following conditions are met:

- the Iceberg extension is loaded;
- `iceberg_warehouse` is configured;
- `file_format='iceberg'`;
- the source matches `namespace.table` or `iceberg_catalog.namespace.table`;

the binder bypasses filesystem globbing and passes the source directly to the Iceberg extension.

Filesystem path validation is also skipped for these Catalog names because they are logical table identifiers, not local files.

### Scope and Compatibility

The change is intentionally limited to Iceberg REST Catalog identifiers.

Existing behavior remains unchanged for:

- Parquet, CSV, JSON, and other file formats;
- regular Iceberg filesystem paths;
- metadata file paths;
- wildcard file patterns;
- REST-like strings when the Iceberg REST configuration is not enabled.

No changes are made to the Iceberg extension implementation itself.

### Result

After rebuilding Ladybug, the following queries can be resolved through the REST Catalog:

```cypher
CALL iceberg_warehouse='warehouse';
CALL iceberg_endpoint='http://127.0.0.1:8181';
CALL iceberg_authorization_type='none';

LOAD FROM 'perf_demo.User'
(file_format='iceberg')
RETURN count(*);
```

```cypher
LOAD FROM 'perf_demo.Session'
(file_format='iceberg')
RETURN count(*);
```

```cypher
LOAD FROM 'perf_demo.User_owns_Session'
(file_format='iceberg')
RETURN count(*);
```

The fully qualified form is also supported:

```cypher
LOAD FROM 'iceberg_catalog.perf_demo.User'
(file_format='iceberg')
RETURN count(*);
```

### DIFF
--- a/src/binder/bind/bind_file_scan.cpp
+++ b/src/binder/bind/bind_file_scan.cpp
@@ -1,3 +1,5 @@
+#include <cctype>
+
 #include "binder/binder.h"
 #include "binder/bound_scan_source.h"
 #include "binder/expression/literal_expression.h"
@@ -120,12 +122,48 @@ bool handleFileViaFunction(main::ClientContext* context, std::vector<std::string
     return handleFileViaFunction;
 }
 
+// Qualified Iceberg REST table names are not filesystem paths and must bypass
+// local globbing before the Iceberg extension resolves them.
+bool isIcebergCatalogName(const main::ClientContext* context, const std::string& source,
+    const case_insensitive_map_t<Value>& options) {
+    // Keep this workaround scoped to a loaded Iceberg REST configuration.
+    auto warehouseOption = context->getExtensionOption("iceberg_warehouse");
+    if (warehouseOption == nullptr ||
+        context->getCurrentSetting("iceberg_warehouse").toString().empty()) {
+        return false;
+    }
+    auto format = options.find(FileScanInfo::FILE_FORMAT_OPTION_NAME);
+    if (format == options.end() || StringUtils::getLower(format->second.toString()) != "iceberg") {
+        return false;
+    }
+    if (source.find('/') != std::string::npos || source.find('\\') != std::string::npos) {
+        return false;
+    }
+    idx_t dot_count = 0;
+    for (idx_t i = 0; i < source.size(); i++) {
+        auto c = static_cast<unsigned char>(source[i]);
+        if (source[i] == '.') {
+            dot_count++;
+            continue;
+        }
+        if ((i == 0 || source[i - 1] == '.') && !std::isalpha(c) && source[i] != '_') {
+            return false;
+        }
+        if (!std::isalnum(c) && source[i] != '_' && source[i] != '$') {
+            return false;
+        }
+    }
+    return dot_count == 1 || dot_count == 2;
+}
+
 std::unique_ptr<BoundBaseScanSource> Binder::bindFileScanSource(const BaseScanSource& scanSource,
     const options_t& options, const std::vector<std::string>& columnNames,
     const std::vector<LogicalType>& columnTypes) {
     auto fileSource = scanSource.constPtrCast<FileScanSource>();
-    auto filePaths = bindFilePaths(fileSource->filePaths);
     auto boundOptions = bindParsingOptions(options);
+    bool catalogName = fileSource->filePaths.size() == 1 &&
+                       isIcebergCatalogName(clientContext, fileSource->filePaths[0], boundOptions);
+    auto filePaths = catalogName ? fileSource->filePaths : bindFilePaths(fileSource->filePaths);
     FileTypeInfo fileTypeInfo;
 
     if (boundOptions.contains(FileScanInfo::FILE_FORMAT_OPTION_NAME)) {
@@ -136,7 +174,7 @@ std::unique_ptr<BoundBaseScanSource> Binder::bindFileScanSource(const BaseScanSo
     }
     // If we defined a certain FileType, we have to ensure the path is a file, not something else
     // (e.g. an existed directory)
-    if (fileTypeInfo.fileType != FileType::UNKNOWN) {
+    if (fileTypeInfo.fileType != FileType::UNKNOWN && !catalogName) {
         for (const auto& filePath : filePaths) {
             if (!LocalFileSystem::fileExists(filePath) && LocalFileSystem::isLocalPath(filePath)) {
                 throw BinderException{std::format("Provided path is not a file: {}.", filePath)};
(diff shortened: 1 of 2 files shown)


adsharma:
The idea sounds good, but the knowledge of an extension (iceberg) leaks into the main repo. How about we design a callback or the extension does this eagerly and sets a bit somewhere for the main repo to understand it shouldn't glob?

ericyuanhui:
> The idea sounds good, but the knowledge of an extension (iceberg) leaks into the main repo. How about we design a callback or the extension does this eagerly and sets a bit somewhere for the main repo to understand it shouldn't glob?

This solution looks much cleaner. I'll go ahead and refactor accordingly. Thank you for your suggestion!

adsharma:
I'll hold the 0.20.2 release for a few hours if you want to get this in.

ericyuanhui:
> I'll hold the 0.20.2 release for a few hours if you want to get this in.

Unfortunately, I am unlikely to meet the timeline. I have been occupied with other urgent tasks and have not yet made the modifications. I apologize for the delay.

ericyuanhui:
The root cause is that the current `LOAD FROM` binding pipeline assumes every file source is a filesystem path.

For a statement such as:

```cypher
LOAD FROM 'perf_demo.User' (file_format='iceberg') RETURN count(*);
```

the flow is currently:

```text
LOAD FROM
  -> Binder::bindFileScanSource()
  -> Binder::bindFilePaths()
  -> VirtualFileSystem::glob()
  -> “No file found that matches the pattern”
  -> Iceberg table function bind callback is never reached
```

The Iceberg extension only receives control later, when Ladybug selects and binds the `iceberg_scan` table function. By that point, the core binder has already attempted local filesystem globbing and rejected `perf_demo.User`.

Therefore, this cannot be fixed solely inside the current Iceberg extension while preserving the existing `namespace.table` / `iceberg_catalog.namespace.table` syntax.

The existing extension mechanisms do not provide a hook before `Binder::bindFilePaths()` is called. In particular:

- `BinderExtension` is intended for binding extension statements, not for classifying `FileScanSource` paths.
- An Iceberg table function bind callback is invoked after path expansion and validation.
- A VFS-based workaround cannot reliably solve this for the current syntax, because VFS routing is based on the path string and does not receive the parsed `file_format='iceberg'` option. Treating every dotted identifier as an Iceberg source would incorrectly intercept ordinary local file paths.

A single global “Iceberg is loaded, do not glob” flag is also insufficient. Iceberg still supports ordinary filesystem-backed tables, metadata paths, and wildcard paths, all of which must retain the normal globbing behavior. Whether a source is a logical REST catalog identifier depends on the individual statement’s source, `file_format`, and current REST catalog configuration—not merely on whether the extension has been loaded.  

@adsharma

adsharma:
```
   source string
      │
      ▼
   extensionFormat?  (file_format declared && FileTypeUtils::fromString → UNKNOWN)
      │ yes → bindFilePaths(paths, passThroughOnNoMatch=true)
      ▼
   VirtualFileSystem::glob(path)          ← globbing is STILL attempted, always
      │
      ├─ matches found (existing dir/file, or wildcard expansion)
      │      → use the expanded paths          (globbing case, works as before)
      │
      └─ empty (nonexistent path, or a logical name like 'default.lineitem')
             → passThroughOnNoMatch: keep the raw string verbatim
      ▼
   getScanFunction → FileType::UNKNOWN → lookup "ICEBERG_SCAN" → extension bind
      ▼
   iceberg bindFuncHelper: catalog-qualified name? → REST query
                           otherwise        → DuckDB iceberg_scan('<source>')
 ```

PR #874, opened 2026-08-31
Add group-key predicate pushdown optimizer

### LINKED ISSUE #873: Feature: [Optimizer] Post-aggregation group-key filter causes Cartesian-product OOM
### API

C++

## Summary

This issue was discovered while executing a multi-metric Cypher query generated by our analytics agent.

To support dynamically selected metrics, the agent composes the query incrementally: each metric is calculated in an independent `MATCH ... WITH` block, and its result is associated with previously calculated metrics through an entity-ID equality predicate such as:

```cypher
WHERE metric_1_entity_id = entity_id
```

This is a valid and convenient pattern for compositional query generation. However, it exposes an optimizer limitation: when the equality predicate appears after an aggregation, it is not pushed below the `AGGREGATE` operator—even when it depends only on grouping keys and can be evaluated safely before aggregation.

As a result, the engine processes a Cartesian product and builds aggregate state for every candidate entity pair before applying the equality predicate.

## Observed impact

The original database contained on the order of tens of thousands of entities, although the exact row count was not captured during the initial investigation.

During execution of the original agent-generated query, macOS Activity Monitor reported that the `lbug` process had grown to approximately **24.27 GB** of memory before the query failed.

The number of candidate entity pairs grows quadratically:

```text
 5,000 ×  5,000 =    25,000,000
10,000 × 10,000 =   100,000,000
20,000 × 20,000 =   400,000,000
30,000 × 30,000 =   900,000,000
```

The `OPTIONAL MATCH` expansion over `EventAPIAccess` may further increase the number of rows processed when each person has multiple events.

To make the issue fast, deterministic, and suitable for regression testing, the reduced reproduction below uses 5,000 `Person` nodes and explicitly limits the buffer pool to 64 MiB with `-d 64`.

The 64 MiB setting is not the default. Without an explicit limit, the buffer pool is derived from the available physical memory. Increasing that limit only delays the failure; it does not address the quadratic plan shape.

## Original agent-generated query

<details>
<summary>Click to expand the full query</summary>

```cypher
MATCH (p:Person)
OPTIONAL MATCH (p)-[:person_has_identity_user]-(iu:IdentityUser)
               -[:user_has_api_access]-(e:EventAPIAccess)
WHERE e.event_time >= CURRENT_DATE() - 29
WITH
    p.id AS entity_id,
    null AS entity_name,
    COALESCE(COUNT(e.id), 0) AS metric_0_value
WITH
    entity_id,
    entity_name,
    metric_0_value

MATCH (p:Person)
OPTIONAL MATCH (p)-[:person_has_identity_user]-(iu:IdentityUser)
OPTIONAL MATCH (iu)-[:user_has_api_access]-(e:EventAPIAccess)
WHERE e.event_time >= CURRENT_DATE() - 29
WITH
    entity_id,
    entity_name,
    metric_0_value,
    p.id AS metric_1_entity_id,
    null AS metric_1_entity_name,
    AVG(e.response_time) AS metric_1_raw_value
WHERE metric_1_entity_id = entity_id
WITH
    entity_id,
    COALESCE(entity_name, metric_1_entity_name) AS entity_name,
    metric_0_value,
    metric_1_raw_value AS metric_1_value
WITH
    entity_id,
    entity_name,
    metric_0_value,
    metric_1_value

MATCH (p:Person)
OPTIONAL MATCH (p)-[:person_has_identity_user]-(iu:IdentityUser)
OPTIONAL MATCH (iu)-[:user_has_api_access]-(e:EventAPIAccess)
WHERE e.event_time >= CURRENT_DATE() - 29
  AND e.event_time <= CURRENT_DATE()
WITH
    entity_id,
    entity_name,
    metric_0_value,
    metric_1_value,
    p.id AS metric_2_entity_id,
    null AS metric_2_entity_name,
    COALESCE(
        SUM(e.request_body_bytes) + SUM(e.response_body_bytes),
        0
    ) AS metric_2_raw_value
WHERE metric_2_entity_id = entity_id
WITH
    entity_id,
    COALESCE(entity_name, metric_2_entity_name) AS entity_name,
    metric_0_value,
    metric_1_value,
    metric_2_raw_value AS metric_2_value
WITH
    entity_id,
    entity_name,
    metric_0_value,
    metric_1_value,
    metric_2_value

MATCH (p:Person)
OPTIONAL MATCH (p)-[:person_has_identity_user]-(iu:IdentityUser)
OPTIONAL MATCH (iu)-[:user_has_api_access]-(e:EventAPIAccess)
WHERE e.event_time >= CURRENT_DATE() - 29
  AND e.event_time <= CURRENT_DATE()
  AND e.http_status_code >= 200
  AND e.http_status_code <= 299
WITH
    entity_id,
    entity_name,
    metric_0_value,
    metric_1_value,
    metric_2_value,
    p.id AS metric_3_entity_id,
    null AS metric_3_entity_name,
    COALESCE(COUNT(e.id), 0) AS metric_3_raw_value
WHERE metric_3_entity_id = entity_id
WITH
    entity_id,
    COALESCE(entity_name, metric_3_entity_name) AS entity_name,
    metric_0_value,
    metric_1_value,
    metric_2_value,
    metric_3_raw_value AS metric_3_value

RETURN
    entity_id,
    entity_name,
    metric_0_value,
    metric_1_value,
    metric_2_value,
    metric_3_value
LIMIT 100;
```

</details>

### 1. Build the shell

```bash
make shell
```

### 2. Create the reproduction script

```bash
cat > /tmp/group_key_predicate_oom.cypher <<'CYPHER'
CREATE NODE TABLE Person(
    id INT64,
    PRIMARY KEY(id)
);

UNWIND range(1, 5000) AS i
CREATE (:Person {id: i});

MATCH (p0:Person)
WITH
    p0.id AS entity_id,
    COUNT(p0.id) AS metric_0_value
MATCH (p1:Person)
WITH
    entity_id,
    metric_0_value,
    p1.id AS metric_1_entity_id,
    AVG(p1.id) AS metric_1_value
WHERE metric_1_entity_id = entity_id
RETURN COUNT(*);
CYPHER
```

### 3. Run it with an explicitly limited 64 MiB buffer pool

```bash
REPRO_DB="$(mktemp -d)/db"

printf ':quit\n' |
    ./build/release/tools/shell/lbug \
        "$REPRO_DB" \
        -d 64 \
        -i /tmp/group_key_predicate_oom.cypher
```

## Actual behavior

After approximately 30 seconds, the reduced query fails with:

```text
Buffer manager exception: Unable to allocate memory! The buffer pool is full and no memory could be freed!
```

## Expected behavior

The query should complete successfully and return:

```text
5000
```

The predicate:

```cypher
metric_1_entity_id = entity_id
```

depends only on grouping keys, so it can be evaluated before the second aggregation without changing query semantics.

## Root cause

The relevant logical plan is effectively:

```text
FILTER(metric_1_entity_id = entity_id)
└── PROJECTION
    └── AGGREGATE BY entity_id, p1.id
        └── CROSS_PRODUCT
            ├── AGGREGATE BY p0.id
            └── SCAN Person AS p1
```

The first aggregate produces 5,000 entity rows. Because the later `MATCH (p1:Person)` introduces a new variable that is not yet connected to `entity_id`, it is combined with all 5,000 persons.

The engine therefore processes:

```text
5,000 × 5,000 = 25,000,000
```

candidate entity pairs and attempts to build aggregate state for up to 25 million groups.

Only after aggregation does the equality predicate reduce the result to the 5,000 pairs where:

```cypher
p1.id = entity_id
```

The later metric stages in the original query repeat the same pattern.

A trailing `LIMIT 100` cannot prevent the failure because the Cartesian product and aggregation must complete before the limit is applied.

If the predicate is written before the aggregation:

```cypher
MATCH (p1:Person)
WHERE p1.id = entity_id
```

the existing filter push-down optimizer can convert the Cartesian product and equality predicate into a hash join, and the query succeeds within the same memory limit.

## Optimization Proposal

Introduce a group-key predicate push-down rule:

1. Detect filters above an aggregate, optionally separated by projections.
2. Split conjunctive predicates.
3. Push a predicate below the aggregate when it is deterministic and composed only of:
   - Aggregate grouping keys
   - Literals
   - Parameters
4. Let the existing filter push-down pass convert a Cartesian-product equality into a hash join.
5. Keep predicates above the aggregate when they reference:
   - Aggregate results
   - Subqueries
   - Random or volatile functions
6. Do not apply the rule to keyless/global aggregates.
7. For left joins, push only probe-side predicates where null-extension semantics are preserved.

The optimized plan should resemble:

```text
AGGREGATE BY entity_id, p1.id
└── HASH_JOIN(entity_id = p1.id)
    ├── AGGREGATE BY p0.id
    └── SCAN Person AS p1
```

This reduces the intermediate cardinality from approximately `O(N²)` to `O(N)` for the entity join.

## Acceptance criteria

- The reduced reproduction succeeds with an explicitly limited 64 MiB buffer pool.
- It returns `5000`.
- The optimized plan no longer contains the Cartesian product below the second aggregate.
- Predicates referencing aggregate values remain above the aggregate.
- Keyless aggregate semantics remain unchanged.
- Random, volatile, and subquery predicates are not pushed unsafely.
- Left-join null-extension semantics are preserved.
- A low-memory regression test covering multiple consecutive metric stages is added.

Mrhs121:
fix https://github.com/LadybugDB/ladybug/issues/873

### DIFF
--- a/src/include/optimizer/group_key_predicate_push_down_optimizer.h
+++ b/src/include/optimizer/group_key_predicate_push_down_optimizer.h
@@ -1,0 +1,49 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <unordered_set>
+
+#include "binder/expression/expression.h"
+#include "planner/operator/logical_plan.h"
+
+namespace lbug {
+namespace optimizer {
+
+// Pushes predicates from a WITH ... WHERE clause below an aggregate when every value referenced
+// by the predicate is a grouping key of that aggregate. In particular, this turns plans shaped as
+//
+//   FILTER(a = b) -> PROJECTION -> AGGREGATE -> ... -> CROSS_PRODUCT
+//
+// into plans where the filter is directly above the CROSS_PRODUCT. The regular filter push-down
+// pass can then rewrite the cross product plus equality predicate into a hash join.
+class GroupKeyPredicatePushDownOptimizer {
+public:
+    void rewrite(planner::LogicalPlan* plan);
+
+private:
+    std::shared_ptr<planner::LogicalOperator> visitOperator(
+        std::shared_ptr<planner::LogicalOperator> op);
+    std::shared_ptr<planner::LogicalOperator> tryRewriteFilter(
+        std::shared_ptr<planner::LogicalOperator> op);
+
+    static std::shared_ptr<planner::LogicalOperator> findAggregateThroughProjections(
+        const std::shared_ptr<planner::LogicalOperator>& op);
+    static bool isGroupKeyOnlyPredicate(const binder::Expression& predicate,
+        const std::unordered_set<std::string>& groupKeyNames);
+    static bool isComposedFromGroupKeys(const binder::Expression& expression,
+        const std::unordered_set<std::string>& groupKeyNames);
+    static bool containsRandomFunction(const binder::Expression& expression);
+
+    std::shared_ptr<planner::LogicalOperator> pushPredicate(
+        std::shared_ptr<planner::LogicalOperator> op,
+        const std::shared_ptr<binder::Expression>& predicate);
+    static std::shared_ptr<planner::LogicalOperator> appendFilter(
+        std::shared_ptr<planner::LogicalOperator> child,
+        const std::shared_ptr<binder::Expression>& predicate,
+        common::cardinality_t cardinality = 0);
+    static void recomputeFlatSchemas(const std::shared_ptr<planner::LogicalOperator>& op);
+};
+
+} // namespace optimizer
+} // namespace lbug
--- a/src/optimizer/CMakeLists.txt
+++ b/src/optimizer/CMakeLists.txt
@@ -9,6 +9,7 @@ add_library(lbug_optimizer
         factorization_rewriter.cpp
         filter_push_down_optimizer.cpp
         foreign_join_push_down_optimizer.cpp
+        group_key_predicate_push_down_optimizer.cpp
         logical_operator_collector.cpp
         logical_operator_visitor.cpp
         optimizer.cpp
--- a/src/optimizer/group_key_predicate_push_down_optimizer.cpp
+++ b/src/optimizer/group_key_predicate_push_down_optimizer.cpp
@@ -1,0 +1,204 @@
+#include "optimizer/group_key_predicate_push_down_optimizer.h"
+
+#include "binder/expression_visitor.h"
+#include "common/enums/accumulate_type.h"
+#include "common/enums/join_type.h"
+#include "planner/operator/logical_accumulate.h"
+#include "planner/operator/logical_aggregate.h"
+#include "planner/operator/logical_filter.h"
+#include "planner/operator/logical_hash_join.h"
+
+using namespace lbug::binder;
+using namespace lbug::common;
+using namespace lbug::planner;
+
+namespace lbug {
+namespace optimizer {
+
+void GroupKeyPredicatePushDownOptimizer::rewrite(LogicalPlan* plan) {
+    plan->setLastOperator(visitOperator(plan->getLastOperator()));
+}
+
+std::shared_ptr<LogicalOperator> GroupKeyPredicatePushDownOptimizer::visitOperator(
+    std::shared_ptr<LogicalOperator> op) {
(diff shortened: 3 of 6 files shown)


adsharma:
Avoiding cartesian product makes a lot of sense and the code looks correct to me. Thank you for the contribution!

PR #878, opened 2026-08-31
fix: re-executed parameterized queries returned stale rows on the cached-plan fast path (#877)

### LINKED ISSUE #877: Bug: re-executed parameterized query returns the first call's rows (joins, sorts, OPTIONAL MATCH, UNION)
### Ladybug version

0.20.0 and 0.20.1. Clean on 0.18.1, 0.19.0 and 0.19.1.

Linux x86-64, Python 3.10–3.14.

### What happened?

Re-executing the same parameterized query string returns the first call's rows.
No error is raised and the rows are well-formed, so callers cannot detect it.
Three nodes with ids 1, 2, 3:

```python
Q = "MATCH (a:N)-[:E]->(b:M) WHERE a.id = $v RETURN b.id"
await conn.execute(Q, {"v": 1})   # 1
await conn.execute(Q, {"v": 2})   # 1 — expected 2
await conn.execute(Q, {"v": 3})   # 1 — expected 3
```

| shape (same three parameters) | 0.19.1 | 0.20.1 |
|---|---|---|
| single-table scan — any parameter type (INT, STRING, `FLOAT[]`); also `IN`, `OR`, `DISTINCT`, `CASE`, `UNWIND`, `WITH` projections, aggregates / `GROUP BY` | `[1, 2, 3]` | `[1, 2, 3]` |
| `ORDER BY` / `SKIP` | `[1, 2, 3]` | `[1, 1, 1]` |
| traversal (1-hop, var-length, shortest path), `OPTIONAL MATCH` | `[1, 2, 3]` | `[1, 1, 1]` |
| `LIMIT`, `UNION ALL` — first call right, later calls return no rows | `[1, 2, 3]` | `[1, None, None]` |

Two observations localise it: `GROUP BY` alone is correct but `GROUP BY` plus
`ORDER BY` is stale, so the sort is the trigger; and a cartesian product of two
scans is correct while a relationship join is not, so it is the join operator,
not touching two tables. `EXISTS { }` subqueries are stale the same way, and
`CALL QUERY_VECTOR_INDEX` / `QUERY_FTS_INDEX` are affected only because they are
top-k.

It is per underlying connection. With `AsyncConnection(max_concurrent_queries=N)`
under concurrent load exactly N calls are correct, one per pooled connection;
sequential awaits reuse one connection, so only the first is. A sync
`Connection` behaves the same.

A `PreparedStatement` is correct, but async callers cannot use one:
`AsyncConnection.execute(PreparedStatement, ...)` raises `UnboundLocalError:
cannot access local variable 'conn_index'`. With separate prepare + execute
deprecated in 0.20.1, and
[the docs](https://docs.ladybugdb.com/get-started/prepared-statements/) telling
Python users to pass parameters straight to `execute()`, the affected form is
the recommended one.

Looks like the sibling of #841, which fixed per-execution state surviving in
cached plans for the aggregate operators — those now pass; sort, hash join,
union, optional-match and subquery appear to carry the same defect. Not
bisected.

**This may already be fixed on main and I could not check.** #870 landed after
0.20.1 was cut, and its second fix — `ResultCollector::prepareForReuse()`
clobbering live results so that "queries returned other queries' rows or empty
tables" — sounds like this. 0.20.1 does not contain it: re-executing
`LOAD FROM df RETURN *` still returns `3, 0, 0` rows there, which is #870's
first symptom. With no nightly after 0.20.1 I have no build of main to test, so
this is reported against the two released versions. If #870 covers it, please
close — the reproducer will confirm either way.

Flagging it because #876 states that cached physical-plan reuse "remains enabled
for read-only parameterized statements", and every shape above is read-only.

Self-contained PEP 723 script — `uv run stale_rows_reexecuted_query.py`
needs nothing but `uv`. Exits 1 when the bug reproduces.

```python
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.10,<3.15"
# dependencies = ["ladybug==0.20.1"]
# ///
"""ladybug 0.20.0/0.20.1: re-executing a parameterized query returns the first
call's rows whenever the plan contains a join, sort, OPTIONAL MATCH, UNION or
subquery.  Clean: 0.18.1, 0.19.0, 0.19.1.

Exits 1 when the bug reproduces.
"""
import asyncio
import os
import sys
import tempfile

import ladybug as kuzu

PARAMS = [1, 2, 3]
CASES = [
    ("scan only",        "MATCH (n:N) WHERE n.id = $v RETURN n.id",                     PARAMS),
    ("ORDER BY",         "MATCH (n:N) WHERE n.id = $v RETURN n.id ORDER BY n.id",       PARAMS),
    ("LIMIT",            "MATCH (n:N) WHERE n.id = $v RETURN n.id LIMIT 1",             PARAMS),
    ("traversal",        "MATCH (a:N)-[:E]->(b:M) WHERE a.id = $v RETURN b.id",         PARAMS),
    ("OPTIONAL MATCH",   "MATCH (n:N) WHERE n.id = $v "
                         "OPTIONAL MATCH (n)-[:E]->(m) RETURN n.id",                    PARAMS),
    ("UNION ALL",        "MATCH (n:N) WHERE n.id = $v RETURN n.id "
                         "UNION ALL MATCH (n:N) WHERE n.id = $v RETURN n.id",           PARAMS),
]


async def main() -> int:
    db = kuzu.Database(os.path.join(tempfile.mkdtemp(), "db"))
    conn = kuzu.AsyncConnection(db, max_concurrent_queries=1)
    await conn.execute("CREATE NODE TABLE N(id INT64, PRIMARY KEY(id));")
    await conn.execute("CREATE NODE TABLE M(id INT64, PRIMARY KEY(id));")
    await conn.execute("CREATE REL TABLE E(FROM N TO M);")
    for i in PARAMS:
        await conn.execute(f"CREATE (:N {{id: {i}}});")
        await conn.execute(f"CREATE (:M {{id: {i}}});")
        await conn.execute(f"MATCH (a:N {{id:{i}}}),(b:M {{id:{i}}}) CREATE (a)-[:E]->(b);")

    print(f"ladybug {kuzu.__version__}")
    stale = []
    for label, query, expected in CASES:
        got = []
        for v in PARAMS:
            result = await conn.execute(query, {"v": v})
            got.append(result.get_next()[0] if result.has_next() else None)
        ok = got == expected
        if not ok:
            stale.append(label)
        print(f"  {'ok   ' if ok else 'STALE'} {label:15} got={got}  expected={expected}")

    if stale:
        print(f"\nBUG REPRODUCED: {', '.join(stale)}")
        return 1
    print("\nclean on this version")
    return 0


sys.exit(asyncio.run(main()))
```

adsharma:
## Summary

Fixes #877.

Re-executing the same parameterized query string — the recommended `execute(query, params)` form, which reuses the cached physical plan — returned the **first execution's rows** whenever the plan contained a sort, top-k, join, cartesian product, `OPTIONAL MATCH`, `UNION`, `EXISTS` subquery, `LIMIT`/`SKIP`, or a recursive extend. The output was well-formed and no error was raised, so callers could not detect it.

Reproducer from the issue (three nodes with ids 1, 2, 3):

```python
Q = "MATCH (a:N)-[:E]->(b:M) WHERE a.id = $v RETURN b.id"
await conn.execute(Q, {"v": 1})   # 1
await conn.execute(Q, {"v": 2})   # 1 — expected 2
await conn.execute(Q, {"v": 3})   # 1 — expected 3
```

| shape (same three parameters) | before | after |
|---|---|---|
| single-table scan | `[1, 2, 3]` | `[1, 2, 3]` |
| `ORDER BY` / `SKIP` | `[1, 1, 1]` / no rows | `[1, 2, 3]` |
| traversal, `OPTIONAL MATCH` | `[1, 1, 1]` | `[1, 2, 3]` |
| `LIMIT`, `UNION ALL` | `[1, None, None]` | `[1, 2, 3]` |

## Root causes

Both are per-execution state surviving across executions of the cached operator tree — the same family as #841 and #870.

### 1. Whole sub-pipelines vanished from the cloned plan template

Several operator `copy()` implementations only cloned `children[0]` and dropped the children the plan mapper attaches afterward (build sides, sort sinks, union collectors). `ProcessorTask::run()` and the cached-plan fast path in `ClientContext::executeNoLock()` both clone through `copy()`, so the corresponding sink pipelines never ran again on later executions:

- `OrderByScan` / `OrderByMerge` / `TopKScan` — ORDER BY and top-k scans lost their sort sinks
- `HashJoinProbe` — dropped the **build side** (traversals, OPTIONAL MATCH, EXISTS, SIP-collector variants)
- `Intersect`, `CrossProduct`, `PathPropertyProbe`
- `UnionAllScan` — UNION / UNION ALL
- `RecursiveExtend`, `TableFunctionCall` (FTable scans), `Profile`, `DummySimpleSink`

Verified empirically: on the second execution of a top-k query the `TOP_K` sink task was never created at all — the sort was simply never re-run.

### 2. Shared states kept per-execution state that was never reset

- `Limit` / `Skip` counters stayed exhausted after the first execution → "first call right, later calls return no rows"
- `HashJoinSharedState` kept the previous execution's rows and hash slots (probe served stale rows)
- `SortSharedState` kept payload tables / sorted key blocks; `strKeyColsInfo` accumulated a duplicate entry per execution; `KeyBlockMergeTaskDispatcher` kept active merge tasks
- `UnionAllScanSharedState` kept the previous scan cursors → no rows after the first call
- `SemiMaskerSharedState` re-merged previous executions' local masks into the global node-offset masks (recursive extend)
- `RecursiveExtendSharedState` kept its limit counter and factorized-table pool contents
- `ResultCollector`: **internal** collectors (union branches, cross-product / accumulate / SIP builds) are only read by other operators of the same plan, which hold references to the same table object — so they must be cleared in place, never replaced. Replacing them with a fresh table (the #870 `use_count` heuristic) left plan-internal readers (FTable scans, `UnionAllScanSharedState`, `CrossProductLocalState`) pointing at the stale table. Only the plan **root** collector — the one whose table is handed to the client via `getQueryResult()` — keeps the `use_count`-based fresh-table behavior for overlapping executions.

Note on parameter values: the cached-plan path already re-reads parameter values live at `resolveResultVector` time (Phase 1 work), so values propagate correctly once the plan structure and state resets are fixed.

## Changes

- `copy()` implementations preserve **all** children
- Per-execution state is reset on hooks that run exactly once per execution:
  - `initGlobalStateInternal()`: `Limit`, `Skip`, `HashJoinBuild`, `UnionAllScan`, `BaseSemiMasker`, `SortSharedState::init()`, `KeyBlockMergeTaskDispatcher::init()`
  - `prepareForReuse()`: `RecursiveExtend`
- `ResultCollector` distinguishes internal vs client-facing result tables (flag propagated through `copy()`)

The `AsyncConnection.execute(PreparedStatement, ...)` `UnboundLocalError` mentioned in the issue does not reproduce on current main (the `conn_index` handling in `async_connection.py` already covers it).

## Testing

- New regression test `ApiTest.RepeatedParameterizedCachedPlanExecution877` (covers ORDER BY, top-k, LIMIT, SKIP, joins, cartesian, OPTIONAL MATCH, UNION ALL, EXISTS, var-length extend, repeated through the cached-plan fast path). **Fails on unpatched main, passes with this change.**
- Full Python suite: 153 passed (includes `test_async_prepare_and_execute_concurrent`, the #870 regression test)
- Full e2e suite: 1961/1962 — the single failure (`dictionary_bug~orb383_relationship_projection_obfuscated.AnonymousParquetDeleteReload`) was verified to fail identically on clean main
- Concurrency: repeated `AsyncConnection` pool stress with overlapping executions and held live results — no corruption, no crashes (an early draft of this change that didn't propagate the internal-table flag through `copy()` segfaulted here; the flag propagation fixes it)

### DIFF
--- a/src/include/common/counter.h
+++ b/src/include/common/counter.h
@@ -17,6 +17,9 @@ public:
 
     bool exceedLimit() const { return counter.load() >= limitNumber; }
 
+    // Re-arm the counter for another execution of a cached physical plan.
+    void reset() { counter.store(0); }
+
 private:
     common::offset_t limitNumber;
     std::atomic<common::offset_t> counter;
--- a/src/include/main/client_config.h
+++ b/src/include/main/client_config.h
@@ -8,6 +8,24 @@
 namespace lbug {
 namespace main {
 
+// Scope of statements for which the cached-physical-plan fast path (see
+// CachedPreparedStatement::physicalPlanCache) may be used when a parameterized prepared
+// statement is re-executed. This is a safety valve for latent state-reuse bugs in the plan
+// cache: setting it to READS, WRITES or NONE trades the optimization for protection.
+enum class CachedPreparedStatementScope {
+    READS = 0,  // cache and reuse plans of read-only statements only
+    WRITES = 1, // cache and reuse plans of write statements only
+    BOTH = 2,   // cache and reuse plans of reads and writes
+    NONE = 3,   // disable plan caching entirely (re-map the plan on every execution)
+};
+
+struct CachedPreparedStatementScopeUtils {
+    // Parses one of [READS, WRITES, BOTH, NONE] (case-insensitive); throws a
+    // RuntimeException otherwise. Defined in settings.cpp.
+    static CachedPreparedStatementScope fromString(const std::string& str);
+    static std::string toString(CachedPreparedStatementScope scope);
+};
+
 struct ClientConfigDefault {
     // 0 means timeout is disabled by default.
     static constexpr uint64_t TIMEOUT_IN_MS = 0;
@@ -24,6 +42,10 @@ struct ClientConfigDefault {
     static constexpr bool ENABLE_PLAN_OPTIMIZER = true;
     static constexpr bool ENABLE_INTERNAL_CATALOG = false;
     static constexpr bool ENABLE_PACKED_PATH_EXTEND = false;
+    // Statement kinds for which the cached-physical-plan fast path is enabled.
+    // BOTH preserves the historical behaviour (all parameterized statements).
+    static constexpr CachedPreparedStatementScope CACHED_PREPARED_STATEMENT_SCOPE =
+        CachedPreparedStatementScope::BOTH;
     // Memory budget (in bytes) for the in-memory primary-key uniqueness buffer used when COPY-ing
     // into a primary-key node table that has no hash index. Once the buffer exceeds this budget it
     // is sorted and spilled to disk as a sorted run; cross-run duplicates are detected during a
@@ -69,6 +91,10 @@ struct ClientConfig {
     // Memory budget (bytes) for the no-hash-index COPY primary-key validator before it spills
     // sorted runs to disk. See ClientConfigDefault::PK_VALIDATOR_SPILL_THRESHOLD.
     uint64_t pkValidatorSpillThreshold = ClientConfigDefault::PK_VALIDATOR_SPILL_THRESHOLD;
+    // Which statement kinds may reuse a cached physical plan when a parameterized prepared
+    // statement is re-executed. See CachedPreparedStatementScope for the safety rationale.
+    CachedPreparedStatementScope cachedPreparedStatementScope =
+        ClientConfigDefault::CACHED_PREPARED_STATEMENT_SCOPE;
 };
 
 } // namespace main
--- a/src/include/main/client_context.h
+++ b/src/include/main/client_context.h
@@ -226,6 +226,9 @@ private:
         CachedPreparedStatement* cachedPreparedStatement,
         std::optional<uint64_t> queryID = std::nullopt, QueryConfig config = {},
         bool cachePhysicalPlan = false);
+    // Whether the cached-physical-plan fast path may be used for the given statement, per the
+    // `enable_cached_prepared_statement` setting.
+    bool isCachedPlanAllowedFor(const PreparedStatement& preparedStatement) const;
     std::unique_ptr<QueryResult> queryNoLock(std::string_view query,
         std::optional<uint64_t> queryID = std::nullopt, QueryConfig config = {});
 
--- a/src/include/main/settings.h
+++ b/src/include/main/settings.h
@@ -166,5 +166,15 @@ struct EnablePackedPathExtendSetting {
     static common::Value getSetting(const ClientContext* context);
 };
 
+struct EnableCachedPreparedStatementSetting {
+    static constexpr auto name = "enable_cached_prepared_statement";
+    static constexpr auto inputType = common::LogicalTypeID::STRING;
+    // One of [READS, WRITES, BOTH, NONE] (case-insensitive): which statement kinds may
+    // reuse a cached physical plan when a parameterized prepared statement is re-executed.
+    // Safety valve for latent plan-cache state-reuse bugs; NONE disables the optimization.
+    static void setContext(ClientContext* context, const common::Value& parameter);
+    static common::Value getSetting(const ClientContext* context);
+};
+
 } // namespace main
 } // namespace lbug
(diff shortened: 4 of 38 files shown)


adsharma:
The test failures are pre-existing in main and tracked by #881

PR #886, opened 2026-09-01
fix: string hash portability across platforms with different char signedness (issue #882)

### LINKED ISSUE #882: Bug: string primary-key index is not portable across platforms with different `char` signedness (non-ASCII keys silently return no rows)
### Ladybug version

0.18.3, 0.19.1, 0.20.1 (identical in all three).

The bug needs two: written on one platform, read on another. Tested macOS arm64, Linux x86-64, Linux aarch64.

### What happened?

A database's **string primary-key index is not portable between platforms where plain `char` has a different signedness.** After moving the file, `=` on a **non-ASCII** primary key silently returns **no rows**. No error: the row is there, `IN [x]` still finds it, ASCII keys are fine, and non-PK string equality is fine. It is symmetric, so neither build is "the broken one".

`char` is unsigned on `aarch64-linux-gnu` and `riscv64-linux-gnu`, signed on `x86_64-linux-gnu`, `arm64-apple-darwin` and `wasm32`. So "build the database on x86-64 CI or a developer Mac, deploy to ARM" silently loses rows for any key with a non-ASCII byte among its last `len % 8` bytes: `José`, `Müller`, Cyrillic, CJK, emoji.

**Cause.** `Hash::operation(const std::string_view&)` in `src/include/function/hash/hash_functions.h`. The main loop reads `uint64_t` blocks and is signedness-independent, but the tail loop indexes the `string_view`, which yields `char`:

```cpp
last |= static_cast<uint64_t>(key[key.size() / 8 * 8 + i]) << i * 8;
```

A signed `char` >= 0x80 sign-extends before the shift, so it also corrupts the higher bytes of `last`. Consistent with this, keys whose byte length is a multiple of 8, which skip the tail loop entirely, are portable.

**Suggested fix:**

```cpp
last |= static_cast<uint64_t>(
    static_cast<unsigned char>(key[key.size() / 8 * 8 + i])) << i * 8;
```

Verified by building the engine twice with `-fsigned-char` as the only difference and cross-reading a database between them: upstream fails in both directions, patched passes in both.

**One decision for you before a PR:** the fix is bit-identical on unsigned-char builds but changes hashes on signed-char ones, so databases already written there need their primary-key index rebuilt. If you would rather bump a storage version, or make the hash explicitly byte-oriented throughout, the one-liner is not the patch you want. Happy to open a PR with a test once you've decided.

Two containers, no data files.

```python
# repro.py
import platform, sys, ladybug as lb
mode, path = sys.argv[1], sys.argv[2]

if mode == "make":
    conn = lb.Connection(lb.Database(path))
    conn.execute("CREATE NODE TABLE T(k STRING PRIMARY KEY, v INT64)")
    conn.execute("CREATE (:T {k: 'mizu', v: 1})")
    conn.execute("CREATE (:T {k: '水', v: 2})")
    for i in range(2000):
        conn.execute(f"CREATE (:T {{k: 'k{i}', v: {i + 10}}})")
    print("built on", platform.machine())
else:
    conn = lb.Connection(lb.Database(path, read_only=True))
    q = lambda s: [r[0] for r in conn.execute(s)]
    print("read on ", platform.machine())
    print("  ascii    =  ->", q("MATCH (t:T) WHERE t.k = 'mizu' RETURN t.v"), "expected [1]")
    print("  nonascii =  ->", q("MATCH (t:T) WHERE t.k = '水' RETURN t.v"), "expected [2]")
    print("  nonascii IN ->", q("MATCH (t:T) WHERE t.k IN ['水'] RETURN t.v"), "expected [2]")
```

```sh
docker run --rm --platform linux/amd64 -v "$PWD:/d" python:3.14-slim \
  sh -c "pip -q install ladybug==0.20.1 && python /d/repro.py make /d/db.lbug"

docker run --rm --platform linux/arm64 -v "$PWD:/d" python:3.14-slim \
  sh -c "pip -q install ladybug==0.20.1 && python /d/repro.py read /d/db.lbug"
```

```
built on x86_64
read on  aarch64
  ascii    =  -> [1] expected [1]
  nonascii =  -> []  expected [2]
  nonascii IN -> [2] expected [2]
```

Swapping the two `--platform` flags reproduces it in the other direction; building and reading on the same platform passes every time.

adsharma:
## Description

Fixes #882.

`Hash::operation(const std::string_view&)` (`src/include/function/hash/hash_functions.h`) hashed the tail bytes (`len % 8`) by indexing the `string_view` directly, which yields plain `char`. The behavior of `char` differs by platform:

- **signed** on x86-64-linux, macOS arm64, wasm32
- **unsigned** on aarch64-linux, riscv64-linux

A byte >= 0x80 in the tail was sign-extended before the shift on signed-char platforms but not on unsigned-char ones, so the same string produced different hashes. A database written on one platform then read on another silently returned no rows for `=` on non-ASCII primary keys (`José`, `Müller`, Cyrillic, CJK, emoji) whenever the key's byte length was not a multiple of 8. Keys of length ≡ 0 (mod 8) skip the tail loop entirely and were always portable — consistent with the report.

## The fix

Standardizes on the **unsigned** behavior (as suggested in the issue, and per review on this PR — PostgreSQL, BerkeleyDB, SQLite and MySQL all use unsigned char semantics by default):

```cpp
last |= static_cast<uint64_t>(static_cast<unsigned char>(key[key.size() / 8 * 8 + i]))
        << i * 8;
```

- On unsigned-char platforms (aarch64/riscv64 linux) this is behaviorally identical to the previous code — **existing databases and index hashes on those platforms remain valid**.
- On signed-char platforms (x86-64, macOS arm64, wasm32) the hash now matches aarch64/riscv64, fixing cross-platform portability.

## Verification

- Compiled a standalone program using this header with `-fsigned-char` and `-funsigned-char`: hashes for `José`, `Müller`, Cyrillic, emoji, and ASCII keys are now identical under both flags.
- Compared against the pre-fix header under `-funsigned-char`: values are byte-identical, confirming no change for existing unsigned-char-platform databases.

## Caveat

Databases *written on signed-char platforms* (x86-64, macOS arm64, wasm32) before this fix have hashes baked in with the sign-extended tail semantics. After this change, reading them on the same platform with an affected non-ASCII key (`len % 8 != 0`) would miss. Affected string-PK indexes need a rebuild/re-import. Worth a note in the release notes.

### DIFF
--- a/src/include/function/hash/hash_functions.h
+++ b/src/include/function/hash/hash_functions.h
@@ -146,7 +146,12 @@ inline void Hash::operation(const std::string_view& key, common::hash_t& result)
     }
     uint64_t last = 0;
     for (size_t i = 0u; i < key.size() % 8; i++) {
-        last |= static_cast<uint64_t>(key[key.size() / 8 * 8 + i]) << i * 8;
+        // Zero-extend the byte explicitly so the hash is identical on platforms where plain
+        // `char` is signed (x86-64, macOS arm64, wasm32) and unsigned (aarch64/riscv64 linux).
+        // This matches the convention of PostgreSQL, BerkeleyDB, SQLite and MySQL, and keeps
+        // behavior identical on unsigned-char platforms (aarch64/riscv64 linux).
+        last |= static_cast<uint64_t>(static_cast<unsigned char>(key[key.size() / 8 * 8 + i]))
+                << i * 8;
     }
     hashValue = lbug::function::combineHashScalar(hashValue, lbug::function::murmurhash64(last));
     result = hashValue;


adsharma:
Postgres: https://www.postgresql.org/message-id/19587-4416a590531c9c73%40postgresql.org

BerkeleyDB, SQLIte and MySQL use unsigned by default. aarch64 and risc-v chose unsigned because that's cleaner. Unsigned seems to be the better option here. But I feel uncomfortable pushing this into an already overloaded 0.20.x queue. 

Let's deal with it in trunk after the 0.20.x patch releases stabilize.

Issue #835, opened 2026-08-24
Bug: Severe performance regression and severe file bloat for COPY FROM on relationship table


ericyuanhui:
### Ladybug version

0.18.1

### What operating system are you using?

ubuntu24

### What happened?

Hello, while running tests today, I accidentally observed a significant performance regression for the `COPY FROM` operation on the relationship table in version 0.19.1. I bisected several historical releases and narrowed down the regression to the changes merged in PR: *Fix O(N²) ALP metadata recomputation in Column*, with commit ID **e179d72f742f4a91a2d79a8243b5a609fd2bfbc9**.

Version | Overall node (1.4M) batch total‑time(s) | Overall node batch total‑rate(rows/s) | Overall rel (0.65M) batch total‑time(s) | Overall rel batch total‑rate(rows/s)
-- | -- | -- | -- | --
0.19.1 | 14.2235 | 98428.97 | 11.3005 | 57519.74
0.18.1 | 14.4803 | 96682.92 | 12.2336 | 53132.41
0.18.0 | 19.9004 | 70350.42 | 7.3231 | 88760.02
0.17.1 | 21.5238 | 65044.42 | 7.6309 | 85180.48

 

After this commit, I also noticed that Ladybug’s graph data files have grown substantially for the same dataset: previously around 150 MB, now reaching 500 MB. The ingestion throughput dropped sharply from 88 000 rows/s down to 53 000 rows/s.

<img width="886" height="134" alt="Image" src="https://github.com/user-attachments/assets/79435058-0a3a-4087-9179-4b323700750f" />

Could you please help analyze whether there are potential improvement options, or if we should revert this change?  @adsharma 

test script on attach file
[LDBC_SAM_TEST.tar.gz](https://github.com/user-attachments/files/31383200/LDBC_SAM_TEST.tar.gz)

### Are there known steps to reproduce?

uv venv --python 3.13 .venv-wheel-test

uv pip install \
  --python .venv-wheel-test/bin/python \
  "ladybug==0.17.1"
  
.venv-wheel-test/bin/python perf_write_SF.py \
  --schema /home/eric/project/LDBC_SAM_TEST/SAM_schema.cypher \
  --copy /home/eric/project/LDBC_SAM_TEST/SAM_copy.cypher \
  --csv-root /home/eric/project/LDBC_SAM_TEST/CsvBasic \
  --db-path /home/eric/project/LDBC_SAM_TEST/lbug_0_17_1.lbug \
[...]

PR #925, opened 2026-09-06
Fix #924: reuse shadow FileHandle across checkpoints

### LINKED ISSUE #924: Bug: each CHECKPOINT permanently consumes ~9 MiB of max_db_size accounting — 7 checkpoints exhaust a 64 MiB cap on a 136 KiB database (0.17.1 and 0.19.0); post-cap checkpoint corrupts the file
*Authorship disclosure: this report was written by Claude, Anthropic's AI, working as the developer on the project that found this behavior; the project's owner reviewed and filed it.*

## Summary

Every `CHECKPOINT` permanently consumes ~9 MiB of the `max_db_size` accounting budget, regardless of how much data was written — including effectively zero. A minimal single-connection loop of tiny-write + `CHECKPOINT` exhausts a 64 MiB `max_db_size` in **exactly 7 checkpoints** with the data file at 136 KiB, failing with:

```
Buffer manager exception: Maximum database size of 67108864 bytes has been reached.
Reopen the database with a larger max_db_size to continue.
```

Reproduced identically on **0.17.1 and 0.19.0** (Python wheels, macOS 15.4 arm64, Python 3.11).

## Why this matters beyond the exception

1. The consumption rate appears constant per checkpoint and independent of buffer pool size: with cognee's shipped defaults (32 GiB cap), a long-lived process hits the wall after ~3,500 checkpoints — and cognee's ladybug adapter checkpoints after every `add_nodes`/`add_edges` call, so a single long ingest run gets there in minutes-to-hours. We hit it twice in production-shaped use.
2. **The state after the cap error is worse than the error:** in both of our incidents, the next `CHECKPOINT` issued on the capped handle crashed with SIGSEGV (`KERN_INVALID_ADDRESS at 0x8`) inside `_lbug` mid-checkpoint, leaving the main DB file permanently unopenable ("The file is not a valid Lbug database file!", with an orphaned `.wal.checkpoint` beside it). That corruption-at-the-fold outcome matches what #713 documents from a different trigger path.
3. Downstream projects have historically responded by raising the cap (e.g. cognee PR topoteretes/cognee#2851 moved their default 4 GB → 32 GiB after "limits don't seem to do much other than raise exceptions") — which defers, rather than fixes, both the exception and the post-cap corruption.

## Reproducer

```python
import ladybug as lb

db = lb.Database("leak-db/db", max_db_size=64 * 1024 * 1024,
                 buffer_pool_size=32 * 1024 * 1024)
conn = lb.Connection(db)
conn.execute("CREATE NODE TABLE IF NOT EXISTS Node(id STRING, payload STRING, PRIMARY KEY(id))")
for i in range(2000):
    conn.execute(
        "MERGE (n:Node {id: $id}) ON CREATE SET n.payload = $p ON MATCH SET n.payload = $p",
        {"id": f"node-{i % 50}", "p": "x" * 200})
    conn.execute("CHECKPOINT;")  # dies at i == 6 (the 7th checkpoint) on 0.17.1 and 0.19.0
```

Create the parent directory first (`mkdir leak-db`). Fails at cycle 7 on both versions; final on-disk size 136 KiB.

## Isolation: checkpoints, not writes

Variant checkpointing every 100 writes fails after **800 writes / 7 checkpoints**; checkpointing every write fails after **8 writes / 7 checkpoints**. Same failure point in checkpoint count, 100× apart in write count. 64 MiB / 7 ≈ 9.1 MiB consumed per checkpoint. With a 32 GiB cap in our production incident the arithmetic matched: ~3,500 checkpoints over a multi-batch ingest before the wall.

## Expected

Checkpointing is a durability operation; issuing it N times on an otherwise small database should not consume the database's address-space budget. Space taken by a fold's working artifacts (shadow/WAL frame groups) should be returned or reused after the fold completes.

## Environment

- ladybug 0.17.1 and 0.19.0 (PyPI wheels), Python 3.11.12, macOS 15.4.1 (Darwin 24.4.0), Apple M3 (arm64)
- Single process, single connection, no concurrency

Happy to run instrumented builds or additional variants if useful.

adsharma:
Fixes #924.

**Root cause:** every CHECKPOINT orphaned the shadowing FileHandle in ShadowFile::reset() (shadowingFH = nullptr + unlink), and the next checkpoint created a fresh one via BufferManager::getFileHandle(). Each FileHandle allocates VMRegion frame groups (1024 pages / ~4 MiB each) that count against max_db_size and are never reclaimed — so a 64 MiB cap died after exactly 7 checkpoints on a ~136 KiB database.

**Fix:**
- ShadowFile::reset() keeps the shadowing FileHandle alive: evict pages, truncate to zero pages (which also truncates + unlinks the on-disk .shadow so recovery probes don't see a stale file), and reuse the same frame groups next checkpoint.
- FileHandle::removePageIdxAndTruncateIfNecessary() keeps frameGroupIdxes/pageCapacity at their high-water mark so truncated page groups reuse frame groups on regrowth.

**Test:** adds SystemConfigTest.testRepeatedCheckpointsDoNotExhaustMaxDBSize (50 write+CHECKPOINT cycles under a 64 MiB cap). Verified it fails before the fix and passes after; all 7 SystemConfigTest tests pass.

### DIFF
--- a/src/include/storage/file_handle.h
+++ b/src/include/storage/file_handle.h
@@ -104,6 +104,9 @@ public:
     common::page_idx_t getNumPages() const { return numPages; }
     common::FileInfo* getFileInfo() const { return fileInfo.get(); }
     void resetFileInfo() { fileInfo.reset(); }
+    void setFileInfo(std::unique_ptr<common::FileInfo> fileInfo_) {
+        fileInfo = std::move(fileInfo_);
+    }
 
     uint64_t getPageSize() const {
         return isLargePaged() ? common::TEMP_PAGE_SIZE : common::LBUG_PAGE_SIZE;
--- a/src/storage/file_handle.cpp
+++ b/src/storage/file_handle.cpp
@@ -72,6 +72,10 @@ page_idx_t FileHandle::addNewPages(page_idx_t numNewPages) {
 page_idx_t FileHandle::addNewPageWithoutLock() {
     if (numPages == pageCapacity) {
         addNewPageGroupWithoutLock();
+    } else if (numPages >= static_cast<page_idx_t>(pageStates.size())) {
+        // Reusing retained capacity after a truncate: re-expose the already-allocated slots.
+        // ConcurrentVector::resize never deallocates, so this is cheap and leak-free.
+        pageStates.resize(pageCapacity);
     }
     pageStates[numPages].resetToEvicted();
     const auto pageIdx = numPages++;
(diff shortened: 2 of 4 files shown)


adsharma:
#835 may also be related. Cc: @ericyuanhui

PR #932, opened 2026-09-07
fix(storage): prevent corrupted HASH PK entries from crashing MERGE

### LINKED ISSUE #931: Bug: prevent corrupted HASH PK entries from crashing MERGE
### Ladybug version

main

ubuntu24.04

## Title

prevent corrupted HASH PK entries from crashing MERGE

## Problem

A database containing a stale or corrupted HASH primary-key (`_PK`) index entry can crash with SIGSEGV during `UNWIND + MERGE`.

The observed core dump shows this call path:

```text
NodeTable::lookupPK
  -> NodeTable::validatePkNotExists
  -> NodeTable::insert

MERGE attempts to insert a new node, so validatePkNotExists() first checks the PK index. The HASH PK index returns an offset which is then used to locate a node group. When that offset points to a missing node group or past the end of a node group, release builds can dereference an invalid/null node group and crash.

_No response_

### PR DESCRIPTION
**Pull request by ericyuanhui**
fix: https://github.com/LadybugDB/ladybug/issues/931 

## Root Cause

The immediate root cause is an invalid offset stored in, or returned by, the persistent HASH PK index.

The lookup path previously assumed that every index offset was valid:

```text
HASH PK lookup
  -> offset
  -> node group lookup
  -> node group visibility check
```

No runtime validation existed between the index result and node-group access. Therefore an invalid index entry became a native invalid-memory access instead of a database error.

The supplied database reproduces this failure even after copying it and executing a single-threaded, one-batch `UNWIND + MERGE`, without running `COPY` in that execution. This proves that the persisted database state already contains the problematic index state.

The exact producer of that invalid index entry cannot be proven from the available database and core dump alone. The historical workflow included:

```cypher
COPY ... (IGNORE_ERRORS=true (DUPLICATE_PK_ONLY))
```

followed by `UNWIND + MERGE`, so COPY-local PK-index/row lifecycle inconsistency is a credible source, but it is not conclusively established.

## Example Impact

Given an index entry such as:

```text
id = "new-id" -> offset = invalid node group / invalid in-group row
```

a statement such as:

```cypher
UNWIND [{id: 'new-id'}] AS row
MERGE (n:Table {id: row.id})
```

must validate whether `new-id` already exists before inserting it. Instead of returning a database error for the invalid index entry, the previous code dereferenced the invalid offset and terminated the process with SIGSEGV.

This affects both writes and reads that perform HASH PK lookup. The most severe effect is process termination and possible interruption of the active transaction.

## Fix

Validate HASH PK lookup results before accessing node-group storage:

- reject `INVALID_OFFSET`;
- reject offsets whose node group does not exist;
- reject offsets beyond the number of rows in the target node group;
- throw `RuntimeException` instructing the user to drop and rebuild `_PK`.

The same protection is applied to local node-table PK lookup.

The normal path only adds constant-time integer bounds checks and a node-group pointer read. It adds no scans, no extra disk reads, no locks, and no change to normal COPY or query execution semantics.

A corrupted entry is deliberately not treated as “PK does not exist”: doing so could permit a duplicate primary key to be inserted.

## Regression Test

Added coverage for the normal historical workflow:

```cypher
COPY ... (IGNORE_ERRORS=true (DUPLICATE_PK_ONLY), PARALLEL=false);
UNWIND [...]
MERGE (...)
SET ...
```

The test verifies that duplicate-PK-only COPY followed by MERGE correctly matches existing rows, inserts a new row, and preserves PK uniqueness.

## Relation to `e98835d3f`

`e98835d3f` is the only previously identified change directly related to the historical workflow: it fixes HASH PK rollback after a failed COPY by discarding the inserted local PK entry.

The supplied core dump proves that a HASH PK lookup later returned an invalid/stale node offset, which was dereferenced during `validatePkNotExists()` and caused SIGSEGV. The available evidence does not prove that this invalid entry was created specifically by the failed-COPY path fixed in `e98835d3f`; it only makes that path a credible source.

This PR closes the remaining safety gap: regardless of how an invalid HASH PK offset entered the index, PK lookup reports a recoverable index-corruption error instead of dereferencing it and crashing the process.

## Limitations

- This patch prevents the crash and preserves PK correctness, but it does not repair an already corrupted `_PK` index.
- Affected databases require dropping and recreating `_PK` (or an equivalent index rebuild operation).
- The original SQL sequence can be tested for correct behavior on a fresh database, but no normal SQL-only reproducer has been found that creates the corrupted index state from scratch.
- Therefore, the test covers the intended COPY + MERGE workflow, not construction of a malformed persisted index file.
- The patch intentionally does not change `DUPLICATE_PK_ONLY` transaction semantics beyond existing fixes, because broadening that behavior would increase scope and may affect COPY performance and error-handling semantics.

### DIFF
--- a/src/storage/local_storage/local_node_table.cpp
+++ b/src/storage/local_storage/local_node_table.cpp
@@ -3,6 +3,7 @@
 #include "catalog/catalog_entry/node_table_catalog_entry.h"
 #include "common/cast.h"
 #include "common/exception/message.h"
+#include "common/exception/runtime.h"
 #include "common/types/types.h"
 #include "common/types/value/value.h"
 #include "storage/index/hash_index.h"
@@ -42,9 +43,19 @@ void LocalNodeTable::initLocalHashIndex(MemoryManager& mm) {
 }
 
 bool LocalNodeTable::isVisible(const Transaction* transaction, offset_t offset) const {
+    if (offset < startOffset) {
+        throw RuntimeException(
+            "Primary-key index contains an invalid local node offset. Please drop and rebuild "
+            "the _PK index.");
+    }
     auto [nodeGroupIdx, offsetInGroup] =
         StorageUtils::getNodeGroupIdxAndOffsetInChunk(offset - startOffset);
-    auto* nodeGroup = nodeGroups.getNodeGroup(nodeGroupIdx);
+    auto* nodeGroup = nodeGroups.getNodeGroup(nodeGroupIdx, true /*mayOutOfBound*/);
+    if (nodeGroup == nullptr || offsetInGroup >= nodeGroup->getNumRows()) {
+        throw RuntimeException(
+            "Primary-key index contains an invalid local node offset. Please drop and rebuild "
+            "the _PK index.");
+    }
     if (nodeGroup->isDeleted(transaction, offsetInGroup)) {
         return false;
     }
--- a/src/storage/table/node_table.cpp
+++ b/src/storage/table/node_table.cpp
@@ -877,6 +877,9 @@ bool NodeTable::isVisibleNoLock(const Transaction* transaction, offset_t offset)
         return false;
     }
     const auto* nodeGroup = getNodeGroupNoLock(nodeGroupIdx);
+    if (nodeGroup == nullptr) {
+        return false;
+    }
     return nodeGroup->isVisibleNoLock(transaction, offsetInGroup);
 }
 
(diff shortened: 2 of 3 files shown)


src/storage/local_storage/local_node_table.cpp
```diff
@@ -42,9 +43,19 @@ void LocalNodeTable::initLocalHashIndex(MemoryManager& mm) {
...
+    if (offset < startOffset) {
+        throw RuntimeException(
+            "Primary-key index contains an invalid local node offset. Please drop and rebuild "
+            "the _PK index.");
+    }
     auto [nodeGroupIdx, offsetInGroup] =
         StorageUtils::getNodeGroupIdxAndOffsetInChunk(offset - startOffset);
-    auto* nodeGroup = nodeGroups.getNodeGroup(nodeGroupIdx);
+    auto* nodeGroup = nodeGroups.getNodeGroup(nodeGroupIdx, true /*mayOutOfBound*/);
+    if (nodeGroup == nullptr || offsetInGroup >= nodeGroup->getNumRows()) {
```
adsharma, line 54: Could you generalize this code in a follow up?

 - The bounds check lives in the wrong layer. The offsetInGroup >= getNumRows() validation exists only in lookupPK's lambda. lookupPKRange,
   lookupIndexRange, and lookupIndex (node_table.cpp) still pass raw isVisibleNoLock as the visibility callback, and isVisibleNoLock itself got only a
   nullptr check — no row-bounds check.
 - Why it matters: NodeGroup::isVisibleNoLock is null-safe at the chunked-group level (returns false when no chunk found), but an OOB row that still
   falls inside a chunked group's [startRow, startRow + capacity) window proceeds to isDeleted/isInserted version-array reads — the same OOB-read crash
   class. A corrupt index entry reached via a range scan can therefore still crash.
 - Recommendation: move the validation into NodeTable::isVisibleNoLock itself (single choke point for all index-scan paths) instead of duplicating it in
   the lookupPK lambda. That also resolves the current policy inconsistency (point lookup throws, range scan silently skips → the thing the PR body says
   must not happen, "treated as PK does not exist," still happens on scan paths).

adsharma:
Thank you for the fix!

src/storage/local_storage/local_node_table.cpp
ericyuanhui: fix on： https://github.com/LadybugDB/ladybug/pull/944 thx

PR #941, opened 2026-09-08
Fix use-after-free: reusable ResultSet outlived its Database

udamaster:
### What goes wrong

Prepared statement → close → reopen → any query crashes the single-threaded Wasm build (null function). A cached ResultSet outlives its database's MemoryManager. Fix: own the cache per database.

`ProcessorTask::run` keeps the reusable `ResultSet` (the "Phase 3" optimisation that avoids
re-allocating DataChunks for repeated prepared-statement executions) in a process-wide
`thread_local`. That `ResultSet` owns `ValueVector`s whose buffers were allocated by the
`MemoryManager` of the database that created it.

Nothing clears the slot when that database is closed. On multithreaded builds the worker threads
are joined when the database's `TaskScheduler` goes away, which takes the `thread_local` with
them, so this stays hidden. On single-threaded builds (`SINGLE_THREADED`, e.g. the
single-threaded Wasm package) the calling thread runs the tasks, and the slot lives on. The next
query on that thread — in a new database, or the same one reopened — has a different descriptor
id, so the old `ResultSet` is released and its buffers are freed through a `MemoryManager*` that
no longer exists. That is a heap-use-after-free. If the descriptor id happens to match instead, the stale `ResultSet` is
reused, and its buffers are written through the same dangling pointer.

On native single-threaded builds this is silent heap corruption. On the single-threaded
WebAssembly build it surfaces as `RuntimeError: null function` or `table index is out of bounds` a few statements
later, when whatever object landed in the corrupted memory (in our case a parsed statement's
`shared_ptr` control block) is destroyed. Whether it crashes at all depends on allocation
history — path length, how many prepared statements ran, etc. — which made it hard to pin down.

Minimal sequence (browser Wasm, one thread):

1. open database, create a node table with STRING columns
2. `prepare` + `execute` any write, e.g. `CREATE (:T {key: $key, value: $value})`
3. close the database
4. reopen it and run any query (`CALL show_tables() RETURN name;` is enough)

AddressSanitizer on the Wasm build reports:

```
heap-use-after-free in MemoryManager::updateUsedMemoryForFreedBlock
  ~MemoryBuffer ← ~BufferBlock ← ~StringAuxiliaryBuffer ← ~ValueVector ← ~ResultSet
  ← ProcessorTask::run   (region freed by Database::~Database)
```

### The fix

Move the cache into the database: `QueryProcessor` now owns a small `ResultSetPool` with one
slot per executing thread (a `std::unordered_map<std::thread::id, …>` behind a mutex). It is
destroyed with the `QueryProcessor`, which `Database` tears down before its `MemoryManager`, so
a cached `ResultSet` can never outlive the allocator that backs it.

Two details:

- `ProcessorTask::run` holds the returned `shared_ptr` for the duration of the task, so a
  nested query on the same thread (e.g. from a table function) can replace the slot without
  pulling the `ResultSet` out from under the outer task.
- Reuse behaviour is unchanged: repeated executions of the same prepared statement on the same
  thread still get the same `ResultSet` after `resetForReuse()`.

The lock is taken once per task run; the cost is negligible next to the 16 KB+ allocation the
cache exists to avoid.

### Test

`SystemConfigTest.testResultSetReuseDoesNotOutliveDatabase` runs prepared writes, closes and
reopens the database on the same thread, and queries again, three times over. It skips in
in-memory mode (nothing to reopen). Under ASan in a `SINGLE_THREADED` build it fails before this
change and passes after; on multithreaded builds, and without ASan, it exercises the reuse path
and checks the data survives the reopen.

### Verified

- Wasm (Emscripten 6.0.9, single-threaded, `-fsanitize=address`): the sequence above and a
  58-operation close/reopen/vector stress produce zero reports with this change; the unpatched
  build reports the use-after-free on the first reopen.
- Release Wasm build: the original failing sequences pass through both the sync and async
  browser APIs.

### DIFF
--- a/src/include/processor/processor.h
+++ b/src/include/processor/processor.h
@@ -1,15 +1,48 @@
 #pragma once
 
+#include <mutex>
+#include <thread>
+#include <unordered_map>
+
 #include "common/task_system/task_scheduler.h"
 
 namespace lbug {
 namespace main {
 class QueryResult;
 }
+namespace storage {
+class MemoryManager;
+}
 namespace processor {
 class FactorizedTable;
 class PhysicalPlan;
 class PhysicalOperator;
+class ResultSet;
+struct ResultSetDescriptor;
+
+// Per-database pool of reusable ResultSets, one slot per executing thread. A cached ResultSet
+// owns ValueVectors whose buffers belong to this database's MemoryManager, so the pool must be
+// owned by (and destroyed with) the database rather than living in a process-wide thread_local:
+// a thread_local slot outlives Database::~Database and later releases (or worse, reuses) buffers
+// through a dangling MemoryManager pointer when the next database runs a query on that thread.
+class ResultSetPool {
+public:
+    // Returns the ResultSet cached for the calling thread when its descriptor id matches,
+    // otherwise allocates a fresh one and caches it. The caller must keep the returned
+    // shared_ptr for as long as it uses the ResultSet: a nested query on the same thread may
+    // replace the slot, and the pool's own reference is dropped when this database closes.
+    std::shared_ptr<ResultSet> getOrCreate(ResultSetDescriptor* descriptor,
+        storage::MemoryManager* memoryManager);
+
+private:
+    struct Entry {
+        uint64_t descriptorID = UINT64_MAX;
+        std::shared_ptr<ResultSet> resultSet;
+    };
+    std::mutex mtx;
+    std::unordered_map<std::thread::id, Entry> entries;
+};
+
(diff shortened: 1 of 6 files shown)


adsharma:
Thank you for the contribution. It looks like a good trade-off in spite of the small increase in cost for the native case:

 - hit: lock → map find → unlock → resetForReuse() outside lock.
 - miss: make_shared<ResultSet> outside lock → lock → swap → unlock.
 
 Previously it paid almost nothing by using a thread local `ResultSet`

PR #944, opened 2026-09-08
Centralize Node Offset Validation in NodeTable::isVisibleNoLock

ericyuanhui:
follow suggestion : https://github.com/LadybugDB/ladybug/pull/932  

Changes:

1. Centralized index offset validation in `NodeTable::isVisibleNoLock`:
   - Rejects `INVALID_OFFSET`.
   - Rejects missing node groups.
   - Rejects offsets where `offsetInGroup >= nodeGroup->getNumRows()`.

2. Invalid index offsets now consistently throw an error instructing users to drop and rebuild the index.

3. Removed the duplicated validation from the `lookupPK` visibility lambda. It now reuses `isVisibleNoLock`.

4. `lookupPKRange`, `lookupIndexRange`, and `lookupIndex` already use `isVisibleNoLock`, so they now receive the same protection automatically.

5. The existing validation in `LocalNodeTable` remains unchanged for transaction-local, uncommitted nodes.

Formatting was checked with the CI clang-format command.

Added a regression test and verified formatting.

The new test in [drop_index_test.cpp](/home/eric/project/ladybug/test/api/drop_index_test.cpp:78) verifies that:

- A node group has allocated capacity.
- It contains only one actual row.
- An offset of `1` is passed, which is beyond `getNumRows()` but may still fall within the chunk's allocated capacity.
- `isVisibleNoLock` throws `RuntimeException` rather than returning `false` or accessing version arrays out of bounds.

### DIFF
--- a/src/storage/table/node_table.cpp
+++ b/src/storage/table/node_table.cpp
@@ -867,6 +867,10 @@ bool NodeTable::isVisible(const Transaction* transaction, offset_t offset) const
 }
 
 bool NodeTable::isVisibleNoLock(const Transaction* transaction, offset_t offset) const {
+    if (offset == INVALID_OFFSET) {
+        throw RuntimeException(
+            "Index contains an invalid node offset. Please drop and rebuild the index.");
+    }
     if (transaction && transaction->isUnCommitted(tableID, offset)) {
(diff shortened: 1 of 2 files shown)


adsharma:
Looks great! Thank you.

Issue #940, opened 2026-09-07
Heap corruption / SIGSEGV on insert + CHECKPOINT into a node table with an HNSW index (0.20.2 and main; 0.19.1 clean) — deterministic synthetic reproducer


fabzter:
## Summary

A single-connection loop of **`CREATE (:Memory {...})` + `CHECKPOINT`** into a node table that carries an **HNSW vector index** (plus an FTS index) crashes **0.20.2** and **main** (`11a991bc`, built from source) within a few dozen cycles, with heap-corruption signatures on a `TaskScheduler` worker thread. **0.19.1 runs the same script clean** (1000 cycles). Fully synthetic, deterministic reproducer below — no real data needed.

Found while retesting #843 (durable corruption) on newer builds: the real database from that issue also triggers this crash on every 0.20-line build, including an `EXPORT`/`IMPORT`-migrated copy, which is why I could not complete that retest. Filing separately because the trigger here is schema + data shape, not the on-disk history of that file.

## Environment

| | |
|---|---|
| ladybug (Python) | **0.20.2** (PyPI, `cp311 macosx arm64`, `_lbug` UUID `b01df503-10c2-34e5-9a8c-c90880092af7`) and **main @ `11a991bc`** (self-built, FTS/VECTOR extensions built from the same commit and loaded by path) |
| control | 0.19.1 (PyPI) — clean |
| Python | 3.11.14 (Homebrew) |
| OS | macOS 26.6.1 (25G76), arm64 |
| Extensions | `INSTALL FTS; INSTALL VECTOR;` (release) |

## Results

`python repro_hnsw_checkpoint_crash.py <workdir> 300` (supervisor script; the soak runs in a child, the last line is the verdict):

| Build | seed 1 | seed 2 | seed 3 | seed 4 | seed 5 |
|---|---|---|---|---|---|
| 0.19.1 | `CLEAN (339 Memory rows)` | `CLEAN` | | | |
| 0.20.2 | `CRASHED (signal 11 SIGSEGV)` | `CRASHED (signal 11 SIGSEGV)` | `CRASHED (signal 5 SIGTRAP)` | `CRASHED (signal 5 SIGTRAP)` | `CRASHED (signal 11 SIGSEGV)` |
| main `11a991bc` | `CRASHED (signal 11 SIGSEGV)` | | | | |

[...]

Issue #843, opened 2026-08-27
Durable on-disk corruption: one 8-byte page-list header overwritten with a float32, making the DB unopenable (0.19.1, macOS arm64)


fabzter:
## Summary

A single-process database became permanently unopenable. I traced it to **8 bytes** at one page-list header, and a targeted patch of those 8 bytes fully restored the database (all 38 nodes, FTS working).

This looks **distinct from #840** — there you concluded the crash leaves the database untouched, and I agree; this is durable on-disk damage. Filing separately so the threads don't get mixed.

## Environment

- `ladybug` **0.19.1** (`ladybug-0.19.1-cp311-cp311-macosx_15_0_arm64.whl`)
- `ladybug-memory` 0.1.8, Python 3.11.14
- macOS 26.6.1 (25G76), arm64
- Single-process access only (embedded in one agent process at a time; engine lock respected)

## Symptom

Every open fails identically — including `read_only=True` and `throw_on_wal_replay_failure=False`:

```
RuntimeError: IO exception: Cannot read from file: ~/.hermes/ladybug.lbdb
fileDescriptor: 3 numBytesRead: 0 numBytesToRead: 4096 position: 4727899947008
```

## Analysis

- `position 4727899947008` = `4096 × 1154272448`, i.e. page index **1,154,272,448 = 0x44CCCCC0**.
- The file is **11,812,864 bytes = 2884 pages**. The requested page is ~400,000× beyond EOF.
- The little-endian byte sequence `C0 CC CC 44` occurs **exactly once** in the file: offset **11,780,096**, which is exactly the start of **page 2876**.
- The tail page-list has a clear, self-consistent structure. Pages 2878 / 2880 / 2882 each begin with `(0xFFFFFFFF, prev_page_index)`:

```
page 2876 @11780096: (1154272448, 3310)   <-- both out of range
page 2877 @11784192: (126, 0)
page 2878 @11788288: (4294967295, 2877)
page 2880 @11796480: (4294967295, 2879)
page 2882 @11804672: (4294967295, 2881)
```

  By that pattern page 2876 must read `(0xFFFFFFFF, 2875)`.

- **The damaged value looks like a float.** `0x44CCCCC0` as IEEE-754 float32 is **1638.3984375**, and `1638.4 = 4096 × 0.4` — i.e. a plausible *page fill ratio* sitting in a `uint32` page-index field. The adjacent field, `3310` (`0xCEE`), is also out of range for a 2884-page file.
[...]

PR #950, opened 2026-09-09
fix(storage): validate allocated page ranges and page indices on the write path

### LINKED ISSUE #948: Bad page index reaches the write path: database file left at 892 GiB apparent size (sparse), silently exhausting disk on any copy/backup (0.19.1, 0.20.2, 0.21.0-dev)
## Summary

On one database, a garbage page index reaches the **write** path: the engine writes at page `233,877,255` and leaves the file at **957,961,236,480 bytes (892.2 GiB) apparent** while only **11.8 MiB is allocated**. The database still opens and queries normally, so nothing looks wrong — until something copies or archives the file and the sparse hole gets materialised.

That is how I found it: a single `shutil.copy2()` of the file filled a 926 GiB volume to 100% (523 MiB free) before failing with `ENOSPC`. A backup tool, `tar`, `cp`, `docker build`, or a container-image layer would do the same.

Reproduced identically on **0.19.1**, **0.20.2** and **main @ `7e91732e` (0.21.0-dev, includes #945 and #946)** — same byte count, same page index, every run.

## Why I am filing this separately from #843

Same family of root cause (a bad page index), but this is the write path, and the consequence is independent and severe: silent disk exhaustion and unusable backups, with a database that reports itself healthy. #946 hardened out-of-bounds **PIP/header reads**; nothing stops the **write** from extending the file. A `page_index < num_pages` (or `<= max_db_size / PAGE_SIZE`) check on the write path would turn this into an actionable error at the moment of damage. Truncating to the recorded page count on open or checkpoint would also cap the blast radius.

## Reproduction

Deterministic on the affected database from #843 (11,812,864 bytes = 2884 pages; `Memory` node table with an FTS index on `content` and an HNSW index on `embedding`; 39 rows). Single process, single connection:

```
LOAD EXTENSION FTS; LOAD EXTENSION VECTOR;
CALL DROP_FTS_INDEX('Memory','memory_content_fts');
CHECKPOINT;
-- then, in a loop: CREATE (:Memory {...}); CHECKPOINT;
```

The process dies with SIGSEGV inside the first ~50 cycles, and the file is left inflated:

| build | result | apparent size | allocated |
|---|---|---|---|
| 0.19.1 (PyPI wheel) | SIGSEGV, file inflated | 957,961,236,480 | 11,841,536 |
| 0.20.2 (PyPI wheel) | SIGSEGV, file inflated | 957,961,236,480 | 11,841,536 |
| main `7e91732e` (0.21.0-dev) | SIGSEGV, file inflated | 957,961,236,480 | 11,841,536 |

Forensics on the resulting file:

- apparent size / 4096 = **233,877,255 pages** (`0x0DF0AF07`); byte offset `957,961,236,480`. Not ASCII, not a plausible float — just a deterministic garbage integer, identical across versions and runs.
- the last non-zero byte is at offset `11,812,640`, i.e. inside the original 2884 pages. Everything past that is hole.
- header magic `LBUG` is intact, and the file still opens read-only and answers `MATCH (m:Memory) RETURN count(m)`.

## What does *not* reproduce it

- A fresh store with the same schema, both indexes, 39 synthetic rows, then `DROP_FTS_INDEX` + 100 insert/`CHECKPOINT` cycles: stays at 4.9 MB.
- `CREATE_FTS_INDEX` / `DROP_FTS_INDEX` on a small fresh database: no inflation.
- `SIGKILL`ing a process with a database open mid-write: no inflation.

So it needs this file's accumulated on-disk state, the same precondition as #843. I am preparing a content-scrubbed copy of that database to share; the raw copy is available privately.

## Audit of this machine

I checked every Ladybug database file present (120 files, apparent vs allocated blocks): only files produced by the code path above are inflated. My live store and all of its backups are 1:1. So this is not a slow leak that has been silently affecting normal operation — it takes the fault above to trigger it, and then it is immediate and total.

## Environment

macOS 26.6.1 (25G76) arm64 · Python 3.11.14 · ladybug 0.19.1 / 0.20.2 / 0.21.0-dev (self-built at `7e91732e`, FTS+VECTOR extensions built from the same commit) · ladybug-memory 0.1.8 · APFS, 926 GiB volume.

fabzter:
Fixes #948.

## Symptom

A single bad page index extends the data file far past its end. On the database from #843 an
11.8 MB file ends up reporting **892.2 GiB** of apparent size while only ~12 MiB is allocated.
The database still opens and queries fine, so nothing looks wrong — until something copies or
archives it and the sparse hole is materialised. That is how I found it: one `shutil.copy2()`
filled a 926 GiB volume to 100% (523 MiB free) before failing with `ENOSPC`. A backup tool,
`tar`, or a container build would do the same.

Reproduced identically on 0.19.1, 0.20.2 and main, always the same byte count.

## Root cause

A page index becomes a file offset on the write path exactly as it does on the read path, but
only reads were validated (#946). Nothing checked the ranges the allocator handed out.

I traced the actual write with a backtrace on an instrumented build. It is not the shadow-replay
path I first suspected:

```
LocalFileSystem::writeFile(offset=957961232384)
  <- CompressedFlushBuffer::operator()(span<const uint8_t>, FileHandle*, const PageRange&, const ColumnChunkMetadata&)
  <- ColumnChunkData::flush(PageAllocator&)
  <- Column::checkpointColumnChunkOutOfPlace(...)  /  StringColumn::checkpointSegment(...)
  <- ColumnChunk::checkpoint(...) <- NodeGroup::checkpointInMemAndOnDisk(...)
  <- NodeTable::checkpoint(...) <- StorageManager::checkpoint(...) <- Checkpointer::checkpointStorage()
```

The range came from `PageManager::allocatePageRange`, which returns whatever
`FreeSpaceManager::popFreePages` yields. That free list is deserialized from the database file, so
a corrupted entry is handed out as if it were real: on this database, page **212,650,604** of a
**2884-page** file. The checkpoint flush then writes there and the file grows to 892 GiB.

This also closes a gap in #843: the guard's message proves that database's free page list is
corrupted, which is the missing link between the page corruption reported there and the file
inflation reported in #948.

## Fix

`PageManager::allocatePageRange` rejects a range that is not inside the data file, naming the
range and the file's page count. This is where a corrupted free list surfaces first, and nothing
shrinks the data file, so a legitimate free entry is always inside it. The existing `DASSERT` on
the extension path is kept and backed by a real check, since `DASSERT` is stripped in release
builds. Zero-page allocations are exempt — they write nothing, and their start index legitimately
sits at the current end of the file. Style and message shape follow #946.

An earlier revision of this PR also guarded the `FileHandle` write paths and both shadow-replay
writes as defense in depth. **I removed those**, because your CI showed the recovery-path bound
was simply wrong: `FlakyCheckpointerTest.RecoverFromCheckpointApplyingShadowFailure` legitimately
replays a shadow page into data page 266 of a file that currently has 265 pages — extending the
data file is exactly what recovery is for, so there is no sound page-count bound available at
that point. I could not prove the spiller's use of `writePagesToFile` safe either. The remaining
change is the one place with an authoritative bound, and it is the place the fault actually
originates.

## Verification

On the #948 reproducer (a content-scrubbed copy of the #843 database, published at
https://github.com/fabzter/ladybug-948-repro so this is reproducible without my private data):

```
Runtime exception: Page allocation from the free page list returned pages
[212650604, 212650605), which are out of bounds for a data file with 2884 pages.
The free page list may be corrupted.
```

The file stays 11.8 MB instead of reaching 892 GiB, and the error identifies the corrupted free
list instead of surfacing as a SIGSEGV.

False-positive checks (an earlier revision of this patch rejected zero-page allocations, which the
first of these caught):

- 2000-cycle insert + `CHECKPOINT` soak on a clean store carrying FTS and HNSW indexes, seeds 1
  and 2: both complete, row counts correct.
- Smoke covering DDL, 3000 inserts with strings up to 3 KB, deletes that free pages, **reuse of
  those freed pages through the free list** (the path this patch guards), and `EXPORT`/`IMPORT`:
  all pass.

I have not run the C++ suite locally — `BUILD_TESTS` was off in my build directory — so CI is the
first full run. I did not add a unit test because provoking this needs a database whose free list
is already corrupt; happy to add a direct unit test of the guards if you would like one.

## Relationship to the other open reports

- **#946** hardened the read path for exactly this class of fault. This is the same check on the write path, in the same style.
- **#949** (open) fixes an unsynchronized `push_back` in `OptimisticAllocator` that corrupts the heap during parallel COPY. It touches only `optimistic_allocator.{h,cpp}`, so there is no overlap with this change, and the two are complementary: #949 removes a cause of corruption, this fails fast when corrupted allocator state would otherwise reach the disk.
- **#940** is unaffected and not masked: it still reproduces on all 5 seeds with this patch applied. Two findings from testing it alongside this work, in case they help there: under a full ASAN build (core plus FTS and VECTOR extensions all instrumented) the #940 workload runs **clean for 500 cycles**, twice, where the release build dies within ~40 — so it is allocator-layout or timing sensitive rather than a deterministic off-by-one; and applying #949 on top does **not** fix it (5/5 seeds still crash), which rules out that race as its cause.

### DIFF
--- a/src/storage/page_manager.cpp
+++ b/src/storage/page_manager.cpp
@@ -1,23 +1,60 @@
 #include "storage/page_manager.h"
 
+#include "common/exception/runtime.h"
 #include "common/uniq_lock.h"
 #include "storage/file_handle.h"
 #include "storage/storage_manager.h"
+#include <format>
 
 namespace lbug::storage {
 static constexpr bool ENABLE_FSM = true;
+
+namespace {
+// An allocated page index is turned straight into a file offset by the write path, so a range
+// that does not belong to this file silently extends it rather than failing (issue #948: page
+// 233,877,254 of a 2884-page database left an 11.8 MB file reporting 892 GiB of apparent size,
+// which only surfaces once a copy or backup materialises the sparse hole). The free list is
+// deserialized from the database file, so a corrupted entry lands here first: fail at the
+// source, where the offending range can still be named.
+//
+// The file does shrink, via FreeSpaceManager::handleLastPageRange ->
+// FileHandle::removePageIdxAndTruncateIfNecessary, but that path truncates to the start of the
+// trailing free range and drops that range instead of re-adding it, and every surviving entry
+// sorts below it, so a legitimate free entry stays inside the file.
+void validateAllocatedPageRange(const PageRange& range, const FileHandle& fileHandle,
+    const char* source) {
+    if (range.numPages == 0) {
+        // A zero-page allocation writes nothing, and its start index legitimately sits at the
+        // current end of the file.
+        return;
+    }
+    const auto numPages = fileHandle.getNumPages();
+    if (range.startPageIdx >= numPages || range.numPages > numPages - range.startPageIdx) {
+        throw common::RuntimeException(
+            std::format("Page allocation from {} returned pages [{}, {}), which are out of "
+                        "bounds for a data file with {} pages. The database file may be "
+                        "corrupted.",
+                source, range.startPageIdx,
+                static_cast<uint64_t>(range.startPageIdx) + range.numPages, numPages));
+    }
+}
+} // namespace
 
 PageRange PageManager::allocatePageRange(common::page_idx_t numPages) {
     if constexpr (ENABLE_FSM) {
         common::UniqLock lck{mtx};
         auto allocatedFreeChunk = freeSpaceManager->popFreePages(numPages);
         if (allocatedFreeChunk.has_value()) {
+            validateAllocatedPageRange(*allocatedFreeChunk, *fileHandle, "the free page list");
             version.fetch_add(1, std::memory_order_relaxed);
             return {*allocatedFreeChunk};
         }
     }
     auto startPageIdx = fileHandle->addNewPages(numPages);
+    // DASSERT alone is stripped in release builds, and this invariant guards a file offset.
     DASSERT(fileHandle->getNumPages() >= startPageIdx + numPages);
+    validateAllocatedPageRange(PageRange(startPageIdx, numPages), *fileHandle,
+        "the page extension path");
     return PageRange(startPageIdx, numPages);
 }
 


fabzter:
Updated: force-pushed a reduced version after your CI caught a genuine defect in my first
revision. Thanks, that is exactly the failure I could not have found locally.

`FlakyCheckpointerTest.RecoverFromCheckpointApplyingShadowFailure` failed with **my** exception:

```
Cannot replay shadow page into data page 266 during recovery replay:
out of bounds for a data file of 265 pages. The shadow file may be corrupted.
```

That test is right and I was wrong: recovery replaying a checkpoint legitimately writes past the
current end of the data file, since extending it is the point. My bound there (the old header's
page count plus the shadow page count) has no sound basis, so I have dropped the shadow-replay
guards and the `FileHandle` write guards entirely — I could not prove the spiller's
`writePagesToFile` path safe either.

What is left is the single guard in `PageManager::allocatePageRange`, which is where the fault
actually originates and the one place with an authoritative bound. It still catches #948 with the
same message and the file stays 11.8 MB. Re-verified after the reduction: the reproducer stops
cleanly, 2000-cycle insert+`CHECKPOINT` soaks with FTS and HNSW indexes pass on two seeds, and the
delete / free-page-reuse / `EXPORT`+`IMPORT` smoke passes.

One other thing from that run, for the record rather than as a claim about my change:
`e2e_test_ddl~partitioned.PartitionedAlterRestrictions` also failed, with `Fatal signal 11` right
after `MATCH (n:Q) RETURN n.id, n.k` on a renamed partitioned table. I do not think it is mine —
the patch only adds a comparison that throws — but I cannot rule it out from here, and it did not
appear in the runs I checked on `main` or on #949. If it reappears on this run with the reduced
patch I will dig into it properly.

fabzter:
@adsharma  for some reason I cannot add you as a reviewer but please serve yourself to check and destroy my PR with your knowledge.

src/storage/page_manager.cpp
```diff
@@ -1,23 +1,56 @@
...
 namespace lbug::storage {
 static constexpr bool ENABLE_FSM = true;
 
+namespace {
+// An allocated page index is turned straight into a file offset by the write path, so a range
+// that does not belong to this file silently extends it rather than failing (issue #948: page
+// 233,877,254 of a 2884-page database left an 11.8 MB file reporting 892 GiB of apparent size,
+// which only surfaces once a copy or backup materialises the sparse hole). The free list is
+// deserialized from the database file, so a corrupted entry lands here first: fail at the
+// source, where the offending range can still be named. Nothing shrinks the data file, so a
```
adsharma: `FileHandle::removePageIdxAndTruncateIfNecessary()` does shrink the data file.

src/storage/page_manager.cpp
```diff
@@ -1,23 +1,56 @@
...
+    if (range.numPages == 0) {
+        // A zero-page allocation writes nothing, and its start index legitimately sits at the
+        // current end of the file.
+        return;
+    }
+    const auto numPages = fileHandle.getNumPages();
+    if (range.startPageIdx >= numPages || range.numPages > numPages - range.startPageIdx) {
+        throw common::RuntimeException(
+            std::format("Page allocation from {} returned pages [{}, {}), which are out of "
+                        "bounds for a data file with {} pages. The free page list may be "
```
adsharma: "The free list .." may not be the problem for every case. Handled by `source`.

adsharma:
Thank you for the contribution! The code looks correct to me. Minor nits - please fix.

src/storage/page_manager.cpp
fabzter: You are right, and I tested it: `numPages` does shrink, and I saw it happen to the data file, not just to zero-page resets.

Instrumented `removePageIdxAndTruncateIfNecessary` and every allocation, then ran a workload built to force it (two tables, ~8000 rows with 200-3000 byte strings, `DROP TABLE` on the later table, then six rounds of insert / delete-every-fifth / `CHECKPOINT`):

```
LBUG-DIAG truncate: numPages 4515 -> 2424      <- data file, non-zero target
LBUG-DIAG truncate: numPages 527 -> 0
LBUG-DIAG truncate: numPages 236 -> 0
... 16 truncations total
LBUG-DIAG alloc from the free page list: [524,526) file=1054 pages
... 1073 allocations, 27 of them from the free list
rows final: 5601
NO GUARD EXCEPTION
```

So my comment was wrong and I have rewritten it. What keeps the guard sound is not that the file never shrinks, it is `handleLastPageRange`:

```cpp
if (pageRange.startPageIdx + pageRange.numPages == fileHandle->getNumPages()) {
    fileHandle->removePageIdxAndTruncateIfNecessary(pageRange.startPageIdx);
} else {
    addFreePages(pageRange);
}
```

It truncates to the start of the trailing free range and **drops that range instead of re-adding it**, and `mergePageRanges` has already emitted every other entry below that start, so the surviving entries all sort below the new end. The shrink and the only entry that could have ended up above it happen together, which is why the 4515 -> 2424 truncation above is followed by hundreds more allocations without a false positive.

The comment now says that explicitly and names both functions, rather than claiming nothing shrinks.

src/storage/page_manager.cpp
fabzter: Agreed, fixed. The last sentence now reads "The database file may be corrupted.", matching the wording you used in the #946 read-path checks, so `source` is the only thing that says where the range came from:

```
Page allocation from the free page list returned pages [212650604, 212650605),
which are out of bounds for a data file with 2884 pages. The database file may be corrupted.
```

and on the extension path the same sentence reads `Page allocation from the page extension path ...`.

Issue #879, opened 2026-08-31
Fix ASAN bugs


adsharma:
We have nightly sanitizer runs now. These are beginner friendly tasks to fix

https://github.com/LadybugDB/ladybug/actions/runs/33396106462/job/99500889127

adsharma:
There are threading bugs that manifest while loading LSQB via:

```
SF=1 ./init-and-load.sh
```

workaround:

```
CALL THREADS=1
```

fabzter:
Triage of the latest nightly run ([34400871133](https://github.com/LadybugDB/ladybug/actions/runs/34400871133), main `5936a18c`), in case it saves @Saiteja64 some time. Two things stand out: **ASan and LeakSanitizer are clean** (0 `ERROR: AddressSanitizer`, 0 `ERROR: LeakSanitizer`), and the ~480 UBSan reports collapse to **11 sites, three of which produce 90% of them**.

| reports | site | what UBSan says |
|---:|---|---|
| 315 | `src/include/function/hash/hash_functions.h:144:54` | load of misaligned address, `const long unsigned int`, needs 8-byte alignment |
| 68 | `src/processor/operator/hash_join/join_hash_table.cpp:190:59` | load of misaligned `hash_t` |
| 60 | `src/common/vector/value_vector.cpp:571:21` | reference binding to misaligned address |
| 25 | `src/processor/result/factorized_table.cpp:277:47` | member access within misaligned address |
| 8 | `src/processor/operator/order_by/order_by_key_encoder.cpp:185:38` | **store** to misaligned address |
| 5 | `src/common/types/value/value.cpp:358` and `:362` | load of misaligned address |
| 2 | `src/processor/operator/ddl/alter.cpp:26:36` | reference binding to **null pointer** (`const ResultSet`) |
| 4 | `value_vector.cpp:626`, `value_vector.cpp:103`, `factorized_table.cpp:571` | reference binding to misaligned address |

## They are nearly all one bug shape

Row-format buffers pack values at whatever offset they land on, and the code then reads or writes
them through a `reinterpret_cast` to an 8-byte type. The top two:

```cpp
// hash_functions.h:144 — a string_view's char* has no 8-byte guarantee
auto data64 = reinterpret_cast<const uint64_t*>(key.data());
for (size_t i = 0u; i < key.size() / 8; i++) {
    auto blockHash = lbug::function::murmurhash64(*(data64 + i));   // <-- unaligned load
```

```cpp
[...]

PR #954, opened 2026-09-09
Fix unaligned string hash access

Saiteja64:
This fixes a UBSAN error in string hashing.

The hash code was reading string data as `uint64_t` blocks, but the string data isn't always 8-byte aligned. This happens with Ladybug's inline short strings and causes UBSAN to abort.

I changed the 8-byte load to use `memcpy` into a local `uint64_t` instead.

Also added a regression test that hashes the same bytes from aligned and intentionally misaligned memory and checks that the result is the same.

Part of #879.

### DIFF
--- a/src/include/function/hash/hash_functions.h
+++ b/src/include/function/hash/hash_functions.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <cstring>
 #include <functional>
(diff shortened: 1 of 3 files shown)


adsharma:
Related: #882 

Could you in a follow up commit add tests that use non-ASCII characters?

PR #956, opened 2026-09-10
Fix Parquet writer data race

Saiteja64:
This fixes a TSan data race in Parquet export.

Multiple worker threads could call `ParquetWriter::flush()` concurrently. The existing mutex only protected `flushRowGroup()`, while `prepareRowGroup()` also modifies shared writer state.

Moved the existing lock to cover the full flush operation.

`ApiTest.PrepareExport` previously reported 3 TSan warnings and now passes cleanly across repeated runs. A full TSan suite run also dropped from 10 race reports to 2, with no remaining Parquet-related races.

Part of #879.


adsharma:
The fix looks correct. Trying to see if we can keep it correct while restoring the write parallelism that existed before.

PR #978, opened 2026-09-16
fix DAC page check

### LINKED ISSUE #403: Bug:  Corrupted PrimaryKeyIndex header causes uncontrolled OOM on database open — should fail gracefully
### Ladybug version

_No response_

_No response_

### What happened?

When I opening a database file, if the on-disk `HashIndexHeaderOnDisk` for a table's primary key index contains corrupted data (e.g., `numEntries` =
  16544419524162413700), the `PrimaryKeyIndex` constructor proceeds to allocate memory based on these garbage values, causing the process to consume 17+
   GB of RAM and get killed by the OS OOM killer. There is no validation or sanity check on the deserialized header values before allocation.

Image

  ## Discussion: Prevention Strategies

  This issue raises a broader question about data integrity guarantees. There seem to be two complementary approaches to prevent this class of problem:

For preventing this class of issue, would it make sense to add protection on both sides:
    - **Write path**: ensure critical metadata (like index headers) can't be left in a corrupted state after an interrupted write
    - **Read path**: add basic sanity checks on deserialized values before allocating memory, so corruption surfaces as a clear error rather than an OOM
   crash

  Ideally both should exist — write-path prevention to minimize corruption risk, and read-path validation as a safety net when corruption does happen
  (disk errors, incomplete file copies, etc.).

  I'm not sure which scenario caused the corruption in my case (could be a killed process during write, or a file transfer issue), but either way the
  current behavior of silently allocating 17 GB and getting OOM-killed makes diagnosis very difficult. A clear error message at load time would have
  saved hours of debugging.

_No response_

ericyuanhui:
fix  https://github.com/LadybugDB/ladybug/issues/403

#### Confirmed Direct Cause

The DAC header page was zeroed out or was not successfully persisted to disk, while the metadata continued to point to that page.

This explains the following sequence:

```text
metadata remains valid
DAC page is entirely zeroed
nextHeaderPage becomes 0
opening the database causes an infinite traversal
```

#### Possible Causes of the Corruption That Have Not Yet Been Fully Proven

The corruption was more likely caused by one or a combination of the following factors:

1. A write failure occurred during checkpoint or shadow apply;
2. The underlying storage layer—such as NFS, iSCSI, Longhorn, or the ext4 journal—lost writes that had already been reported as successful;
3. After the checkpoint failed, part of the index state had already been committed in memory, while rollback did not fully restore it;
4. The application treated “the transaction committed, but the checkpoint failed” as an ordinary transaction failure and retried it;
5. The database was copied online using `copy2()`, producing an inconsistent snapshot;
6. Repeated checkpoints, concurrent `COPY` operations, or anomalies in the free-page list or allocator amplified the problem.

### 3. A Significant Risk in the Current Code That Requires Focused Verification

The current [hash_index.cpp (line 75)](/home/eric/project/ladybug/src/storage/index/hash_index.cpp:75) first merges local index changes into the persistent index structure.

Then [PrimaryKeyIndex::checkpoint() (line 703)](/home/eric/project/ladybug/src/storage/index/hash_index.cpp:703) performs:

```text
HashIndex::checkpoint()
write the header
write the DiskArray
flush dirty pages
checkpointInMemory()
```

`checkpointInMemory()` clears `localStorage`, as shown in [hash_index.cpp (line 100)](/home/eric/project/ladybug/src/storage/index/hash_index.cpp:100).

However, the checkpoint of the entire database has not yet completed at this point. The following operations may still fail:

```text
shadow flush
WAL sync
main-file apply
main-file fsync
database header write
```

In this situation, [rollbackCheckpoint() (line 683)](/home/eric/project/ladybug/src/storage/index/hash_index.cpp:683) primarily restores the in-memory state of slots and headers. However, pending changes in `localStorage` that have already been cleared may not be fully recoverable.

This has not been proven to be the cause of the current corrupted file, but it is the most well-supported root-cause candidate in the current code.

In addition, the source code of [NodeTable::checkpoint() (line 755)](/home/eric/project/ladybug/src/storage/table/node_table.cpp:755) explicitly states that hash-index checkpointing uses the live index state rather than the complete `snapshotTxn` view.


ericyuanhui:
However, I have not yet identified the root cause of why this file contains a page whose DAC value is `0`, and I have been unable to reproduce the issue. I only have one corrupted 6 GB `.lbug` database file, so the analysis can only be performed retrospectively through static analysis of the existing environment. I hope this is understandable.

adsharma:
Approved. Thanks for the contribution. A couple of follow-ups:

## 1. Validate HashIndexHeaderOnDisk on load
#978 checks DAC chain/locations but not `HashIndexHeaderOnDisk` values.
`HashIndexHeader` ctor does `1ull << currentLevel` unchecked (UB if >=64);
garbage `numEntries` (e.g. #403's 16544419524162413700) stored as-is.
Fix: fail fast in `PrimaryKeyIndex` ctor with corruption error, e.g.
`currentLevel < 64`, `nextSplitSlotId < (1u << currentLevel)`.
Files: `storage/index/hash_index_header.h`, `storage/index/hash_index.cpp`.

## 2. Regression test for corrupted DAC chain
No test locks in #978 (zero-page, self-loop, cycle, OOB `nextHeaderPage`,
`numHeaders > NUM_HEADERS_PER_PAGE`, `getDiskArray(idx >= numHeaders)`).
Fix: unit test over small/fake FileHandle asserting `RuntimeException`,
not hang/OOM/wild read.
Files: `storage/disk_array_collection.cpp`, `include/storage/disk_array_collection.h`.

## 3. Bound DAC chain length by numPages
Cycle/OOB/page-0 checks terminate, but distinct in-bounds chain can still
run up to `numPages` steps (~1.5M for 6GB), 2x~4KB alloc per step.
Fix: throw if `headersForReadTrx.size() > fileHandle.getNumPages()`.
Files: `storage/disk_array_collection.cpp`.

ericyuanhui:
> Approved. Thanks for the contribution. A couple of follow-ups:
> 
> ## 1. Validate HashIndexHeaderOnDisk on load
> #978 checks DAC chain/locations but not `HashIndexHeaderOnDisk` values. `HashIndexHeader` ctor does `1ull << currentLevel` unchecked (UB if >=64); garbage `numEntries` (e.g. #403's 16544419524162413700) stored as-is. Fix: fail fast in `PrimaryKeyIndex` ctor with corruption error, e.g. `currentLevel < 64`, `nextSplitSlotId < (1u << currentLevel)`. Files: `storage/index/hash_index_header.h`, `storage/index/hash_index.cpp`.
> 
> ## 2. Regression test for corrupted DAC chain
> No test locks in #978 (zero-page, self-loop, cycle, OOB `nextHeaderPage`, `numHeaders > NUM_HEADERS_PER_PAGE`, `getDiskArray(idx >= numHeaders)`). Fix: unit test over small/fake FileHandle asserting `RuntimeException`, not hang/OOM/wild read. Files: `storage/disk_array_collection.cpp`, `include/storage/disk_array_collection.h`.
> 
> ## 3. Bound DAC chain length by numPages
> Cycle/OOB/page-0 checks terminate, but distinct in-bounds chain can still run up to `numPages` steps (~1.5M for 6GB), 2x~4KB alloc per step. Fix: throw if `headersForReadTrx.size() > fileHandle.getNumPages()`. Files: `storage/disk_array_collection.cpp`.

add on https://github.com/LadybugDB/ladybug/pull/982  thx

Issue #979, opened 2026-09-16
Add cast bounds tests for FLOAT and DOUBLE


jrjsmrtn:
`test/test_files/cast/cast_bounds.test` covers the integer types, but nothing pins what a string cast to FLOAT or DOUBLE does at the edges: the largest finite value, the smallest denormal, and out-of-range values, which currently yield `inf`, `-inf` or `0`.

That behaviour comes from the bundled fast_float. Newer fast_float releases report out-of-range input as an error instead, so an upgrade would change `CAST('1e39' AS FLOAT)` from `inf` to a failure with no test noticing.

I have the cases written and passing against 0.20.4 (12 added statements, one file). Happy to open the PR if you want it.

adsharma:
Yes, a PR would be great. Also we need to update fast_float at a regular cadence.

PR #980, opened 2026-09-16
test: add cast bounds cases for FLOAT and DOUBLE

jrjsmrtn:
`cast_bounds.test` pins the integer bounds but nothing for FLOAT or DOUBLE. These cases add the largest finite value, the smallest denormal, and out-of-range strings, which cast to `inf`, `-inf` and `0`.

Why it matters: that behaviour comes from the bundled fast_float. Newer releases report out-of-range values as an error, so an upgrade would silently change `CAST('1e39' AS FLOAT)` from `inf` to a failure. DuckDB casts the same way, so this pins inherited behaviour rather than asserting a bug. That also speaks to updating fast_float at a regular cadence, as mentioned in #979.

Tests only, no source changes. All 7 groups in `cast_bounds.test` pass locally against 0.20.4 on macOS arm64, including the two new ones.

Refs #979

Cases drafted with AI assistance and run locally as described above; I can explain and defend every line.


adsharma:

PR #983, opened 2026-09-17
Fix aggregate binding over unresolved parameter placeholders

### LINKED ISSUE #962: SUM(CASE WHEN <timestamp-col> >= TIMESTAMP($param) ...) fails to bind: SUM received ANY
## Description

`SUM` over a `CASE` whose condition compares a TIMESTAMP column against `TIMESTAMP()` applied to a **bound parameter** fails at bind time with `SUM received ANY`. The same query binds and runs on release 0.20.4, so this is a regression on main.

## Repro (self-contained, in-memory)

```python
import ladybug as lb
db = lb.Database(':memory:')
conn = lb.Connection(db)
conn.execute('CREATE NODE TABLE P (id INT64, ts TIMESTAMP, PRIMARY KEY(id))')
conn.execute('CREATE (:P {id: 1, ts: TIMESTAMP("2021-01-01")})')
conn.execute(
    'MATCH (p:P) WITH CASE WHEN p.ts >= TIMESTAMP($s) THEN 1 ELSE 0 END AS v RETURN SUM(v)',
    {'s': '2020-01-01 00:00:00'},
)
```

Fails on main with:

```
RuntimeError: Binder exception: Function SUM did not receive correct arguments:
Actual:   (ANY)
Expected: DISTINCT (INT128) -> INT128
          (INT128) -> INT128
          ... (numeric overloads only, no ANY)
```

## Bisection (all on the same main build)

| Query shape | Result |
|---|---|
| `RETURN SUM(CASE WHEN 1=1 THEN 1 ELSE 0 END)` | OK |
| `RETURN SUM(CASE WHEN TIMESTAMP($s) >= TIMESTAMP('2020-01-01') THEN 1 ELSE 0 END)` (param, no column) | OK |
| `MATCH (p:P) ... RETURN CASE WHEN p.ts >= TIMESTAMP($s) THEN 1 ELSE 0 END` (column + param, no SUM) | OK |
| `MATCH (p:P) ... WITH CASE ... AS v RETURN SUM(v)` (column + param + SUM) | **FAIL** |
| `MATCH (p:Post) ... SUM(CASE WHEN p.creationDate >= TIMESTAMP('literal') ...)` (column + literal + SUM) | OK |

So the trigger is specifically: aggregate over a `CASE` whose condition mixes a **stored column** with a **parameter-dependent** `TIMESTAMP()` call. The full Q4-shaped query (MATCH over Person/knows/Post/Tag with `WITH DISTINCT`, two such CASEs, `SUM` + GROUP BY) fails the same way.

## Expected

Binds and runs as on 0.20.4 (verified: the LDBC SNB Q4-shaped query returns 10 rows there).

## Versions

- Broken: main @ `ab5f1a76` (Extend hash-join SIP with rel-scan semi masks), local release build, Python (`src_py` @ main) + prebuilt-0.20.4-compatible engine reporting storage v47 max
- Working: release 0.20.4 (PyPI wheel)

Found while running the LDBC SNB Interactive complex-14 suite (graph-benchmark-ldbc PR #17), where this blocks Q4.

YuEfSaEDU:
Statements prepared without parameter values (the C API flow) defer parameter binding: expressions referencing a parameter are bound as ANY-typed placeholders and the statement is re-bound with the supplied values before execution. Aggregate overload matching required exact input types, so an aggregate consuming such a placeholder through a WITH alias — e.g. `SUM` over `CASE WHEN p.ts >= TIMESTAMP($s) THEN 1 ELSE 0 END` — failed to prepare with "Function SUM did not receive correct arguments: (ANY)", a regression relative to 0.20.4, whose bindings always prepared with parameters available.

When no exact overload matches and every input type is unresolved ANY, we now select a deterministic arity- and distinctness-matching overload and defer validation to the mandatory re-bind; the provisional plan is never executed because statements with unknown parameters always re-bind. New regression tests in `test/api/prepare_test.cpp` cover the failing column+param+aggregate shape plus the literal and direct-aggregate controls from the issue's bisection table.

Fixes #962


adsharma:

PR #993, opened 2026-09-18
fix STRING dictionary offsets can underflow and crash the engine

### LINKED ISSUE #992: Bug: STRING dictionary offsets can underflow and crash the engine
### Ladybug version

main

ubuntu24.04

### Problem

A persisted STRING dictionary can contain non-monotonic offsets. For example:

```text
offset[2] = 219
offset[3] = 174
```

The current code relies on a debug-only assertion:

```cpp
DASSERT(offset[next] >= offset[current]);
return offset[next] - offset[current];
```

In Release / RelWithDebInfo builds, `NDEBUG` disables `DASSERT`. Therefore:

```text
174 - 219 == -45
```

is converted to an unsigned length:

```text
0xffffffffffffffd3
```

This invalid length is used to construct a `std::string_view`. A later hash, copy, or comparison can dereference invalid memory and terminate the process with SIGSEGV.

Observed checkpoint path:

```text
COPY studentprofile_has_taglabel
  -> COMMIT / CHECKPOINT
  -> CSRNodeGroup::checkpointColumnInRegion
  -> StringChunkData::appendStringColumnChunk
  -> DictionaryChunk::appendString
  -> std::_Hash_bytes
  -> SIGSEGV
```

The failing table was `studentprofile_has_taglabel`. The crash happened while checkpoint was rewriting existing persistent CSR data, not while parsing the newly imported Parquet row.

_No response_

ericyuanhui:
fix https://github.com/LadybugDB/ladybug/issues/992

### Root cause and proposed fix

The direct root cause is invalid persisted dictionary metadata: string offsets must be monotonic and within the dictionary data size, but this invariant was only protected by `DASSERT`.

The original event that corrupted the dictionary cannot be proven from the crash dump alone. It may be caused by a previous failed checkpoint, a historical concurrency/lifetime bug, or an earlier storage corruption. However, the current code path converts corrupt metadata into an uncontrolled native crash.

Proposed code changes:

- Validate dictionary indexes before dereferencing the offset table.
- Validate that `startOffset <= endOffset <= stringDataSize`.
- Validate persistent dictionary offsets while materializing a dictionary from disk.
- Validate offsets in partial scans and checkpoint materialization paths.
- Throw `StorageException` on invalid metadata instead of constructing an invalid `string_view`.

Relevant modified files:

```text
src/storage/table/dictionary_chunk.cpp
src/include/storage/table/dictionary_chunk.h
src/storage/table/dictionary_column.cpp
src/include/storage/table/dictionary_column.h
```

This does not repair an already-corrupted database. Rebuilding secondary indexes is insufficient because the dictionary is part of the physical STRING column storage. The affected table must be restored from a healthy backup or rebuilt from the original source data.


adsharma (commented):
## Review: correct direction, fail-closed on corrupt metadata

The fix addresses the root crash from #992 correctly: persisted `STRING` dictionary offsets were only guarded by `DASSERT`, so a non-monotonic pair like `offset[2]=219, offset[3]=174` underflows to `0xffffffffffffffd3` in release builds and poisons a `string_view`. Converting all four dereference paths to `throw StorageException` instead of crashing is the right call, and this does not claim to repair already-corrupt DBs (restore/rebuild documented in the description).

### What looks good
- `DictionaryChunk::getStringLength`: index-bounds check + `validateStringRange(start,end)` covers both middle entries and the last-entry `dataSize` sentinel. Ordering is right (validate before `end-start`).
- `DictionaryColumn::scan(state, dictChunk)` (disk materialization): validates the freshly scanned offsets *before* rebasing by `initialDictDataSize`, with `requireFirstOffsetZero=true`. The overflow guard (`max - initialDictDataSize`) is a nice touch since `validateOffsets` alone only bounds by `dataMetadata.numValues`.
- `scanOffsets`: bounds guard before `index + numValues` means the later `< numValues` comparison cannot wrap. Validating the array *including* the synthesized `offsets[numValues]=dataSize` sentinel is correct.
- Partial-scan (`scan`) + `materializeToStringChunkDictionary`: per-pair `end<start || end>dataSize` checks are good defense-in-depth on top of `scanOffsets` range validation. The `start>dataSize` case is indirectly covered (either `end>=start` implies `end>dataSize`, or `end<start` fires), but see nit below.
- `[[unlikely]]`, distinct messages per failure site — good for triage.

### Nits / follow-ups (non-blocking)
1. `getString` re-reads `startOffset` after `getStringLength` already read+validated it. Harmless (no mutation between), but reuses a second `getValue` call. Consider having `getStringLength` return the pair or keeping the validated `startOffset`.
2. `scanOffsets` throws on `numValues==0`, while `validateOffsets` treats `numValues==0` as a no-op early return. No current caller passes 0 (`last-first+1 >= 1`), but `if (numValues==0) return;` in `scanOffsets` would be more consistent than throwing.
3. Per-pair checks only test `endOffset`. Correct by transitivity today, but an explicit `startOffset > dataSize` would make the invariant obvious and robust to future refactors.
4. Pre-existing `DictionaryChunk::sanityCheck()` is a tautology (`offsetChunk->getNumValues() <= offsetChunk->getNumValues()`, always true). This PR was a natural place to wire the new validators in there.
5. `DictionaryChunk::deserialize` still materializes without validation — corruption only surfaces lazily on first `getString*` (e.g. during checkpoint hash, as in #992's stack). Worth considering fail-fast validation on deserialize, or at least documenting the lazy-throw contract.
6. No tests in this PR — tracked separately (unit test for non-monotonic / out-of-range / out-of-bounds index → `StorageException`).

No correctness blockers found. The `Column::scanSegment(..., outputChunk, 0, n)` append semantics confirm the `scannedOffsets = getData() + initialDictSize` arithmetic is correct (it appends after `startLength`), so validation sees exactly the newly scanned slice.

PR #994, opened 2026-09-18
Add DictionaryChunk offset-validation unit tests (covers #992)

adsharma:
## Summary

Unit tests for the STRING dictionary offset corruption described in #992, fixed by #993.

New `dictionary_offset_validation_test` (`test/storage/dictionary_offset_validation_test.cpp`, registered in `test/storage/CMakeLists.txt`) builds a small `DictionaryChunk` (`"hello"` / `"world"`, offsets `[0, 5]`, dataSize 10), corrupts the offset table, and asserts fail-closed behavior:

- `ValidDictionaryReadsSucceed` — sane dictionary still reads back correctly.
- `NonMonotonicOffsetsThrowInsteadOfUnderflowing` — `[0, 5]` → `[8, 3]` (mirrors `offset[2]=219, offset[3]=174` from #992); `getStringLength`/`getString` must throw `StorageException` instead of producing length `0xffffffffffffffd3`.
- `OutOfRangeOffsetThrows` — last offset past `dataSize` throws.
- `OutOfBoundsIndexThrows` — index `>= numOffsets` (incl. `uint32 max`) throws.
- `EmptyDictionaryIndexThrows` — documents the behavior change: empty dict no longer returns `0`, it throws.
- `CorruptDictionaryHashThrowsInsteadOfSegfault` — re-appending a known duplicate routes through `indexTable.find` → `getString` on corrupt bytes (the checkpoint crash path in #992) and must throw rather than segfault. Uses a duplicate key deliberately: a fresh key may hash to an empty bucket and take the append path without touching corrupt entries.

## Dependency

Depends on #993 — this PR is test-only against `main` and its new `EXPECT_THROW` cases fail until #993 merges (they fail safely: the old code returns a bogus length / constructs the view without dereferencing, so no crash in the test itself). Verified all 6 tests pass on top of #993's head (`25932997f`), and existing `string_chunk_scan_test` still passes.

## Test plan

- `ninja -C build/release dictionary_offset_validation_test && ./build/release/test/storage/dictionary_offset_validation_test` — 6/6 pass (with #993).
- `./build/release/test/storage/string_chunk_scan_test` — 5/5 pass.


adsharma:
Landed via #997

Issue #998, opened 2026-09-19
Bug: ALP DOUBLE exception chunk can reach invalid state and crash during checkpoint


ericyuanhui:
### Ladybug version

main release version

### What operating system are you using?

ubuntu24.04

### Problem

A separate crash was observed in:

```text
InMemoryExceptionChunk<double>::finalizeAndFlushToDisk
```

The core showed an invalid ALP exception chunk / invalid internal `ColumnChunkData` state during checkpoint finalization.

Relevant path:

```text
COPY / COMMIT
  -> StorageManager::checkpoint
  -> Column::checkpointSegment
  -> InMemoryExceptionChunk<double>::finalizeAndFlushToDisk
  -> SIGSEGV
```

The ALP exception code uses several debug-only checks:

```cpp
DASSERT(exceptionCount < exceptionCapacity);
DASSERT(exceptionIdx < exceptionCount);
DASSERT(exceptionPageOffset == (page_idx_t)exceptionPageOffset);
```

These checks are removed in Release and RelWithDebInfo builds. If ALP metadata is inconsistent, an exception index is invalid, or the exception buffer is full, the engine may read or write outside valid memory. This can corrupt memory before the final checkpoint crash becomes visible.

PR #999, opened 2026-09-19
fix ALP DOUBLE exception chunk can reach invalid state

ericyuanhui:
fix ：https://github.com/LadybugDB/ladybug/issues/998

### Root cause and proposed fix

The immediate cause is an invalid ALP exception chunk state when checkpoint finalization accesses it.

The core dump proves that the object state was already invalid, but it does not prove the exact first write that corrupted it. Possible causes include:

- `exceptionCount > exceptionCapacity`;
- out-of-range exception reads or writes;
- exception buffer capacity exhaustion during ALP compression;
- invalid ALP metadata or an invalid exception-page range;
- an independent use-after-free or memory overwrite.

Proposed code changes:

- Make `SegmentState::getExceptionChunk<T>()` validate the variant type and reject null chunks at runtime.
- Validate ALP metadata before constructing an in-memory exception chunk:
  - compression type is ALP;
  - ALP metadata exists;
  - `exceptionCount <= exceptionCapacity`;
  - required exception pages fit inside the column chunk page range.
- Add runtime bounds checks to `addException`, `getExceptionAt`, `writeException`, and `removeExceptionAt`.
- Check exception-buffer capacity in `FloatCompression::compressNextPageWithExceptions`.
- Prevent unsigned underflow in `FloatCompression::getNumDataPages`.
- Throw `StorageException` instead of continuing with invalid metadata or out-of-range memory access.

Relevant modified files:

```text
src/include/storage/table/column_chunk_data.h
src/include/storage/table/in_memory_exception_chunk.h
src/storage/table/in_memory_exception_chunk.cpp
src/storage/compression/float_compression.cpp
```

These checks convert known invalid states into controlled query/transaction failures. They cannot fully guarantee safety if another unrelated component has already corrupted arbitrary heap memory. An ASan/UBSan reproduction is still needed to identify the first invalid write or use-after-free, if one exists.


adsharma:
A few follow-ups:

 1. Unhardened paths remain — still raw metadata.floatMetadata() (which is dynamic_cast_checked → bare
    reinterpret_cast in release, see common/cast.h:11):
     - float_compression.cpp:177 decompressFromPage, :203 setValuesFromUncompressed, :241 canUpdateInPlace
     - column.cpp:522-523 checkpoint exceptionCount copy
     - compression.cpp:454 toString, column_chunk_metadata.cpp:225 (latter is locally-constructed metadata, low
       risk). Decompress is the most reachable-from-disk path and should call getValidatedMetadata like the compress
       path does.
 2. safeIntegerConversion is DASSERT-only (common/utils.h:51). The constructor's
    safeIntegerConversion<page_idx_t>(numPagesFromExceptions(...)) is unchecked narrowing in release. It happens to
    be guarded by the earlier numPages > metadata.getNumPages() check, but that's implicit — use the explicit
    static_cast-after-range-check pattern you used in column_reader_writer.cpp.
    
 ## Maintainability

 - Validation is now duplicated in FloatCompression::validateALPMetadata and
   InMemoryExceptionChunk::validateMetadata (child-count/type checks). The latter could delegate to the former plus
   its extra page-range checks.
 - Constructor calls getValidatedMetadata(state) 3× (2× in init list); cache once.
 - finalize(state, floatMetadata) takes a const ref but then re-fetches a mutable pointer via unchecked
   floatMetadata() to update exceptionCount — pass mutable or document why the ref stays valid.
   
 ## Security

 - Residual: CompressionMetadata::getChild is also DASSERT-only (compression.cpp:125-128); any unvalidated caller
   (the decompress paths above, plus public FloatCompression::numValues/getBitpackInfo) is still an OOB vector in
   release on crafted metadata.
 - Residual (pre-existing): no cap on exceptionCapacity beyond page-range fit (capacity ≤ numPages × ~341 still
   allows ~1e12-entry NullMask/buffer alloc from a crafted file → OOM). Worth a follow-up bound, not this commit.**

PR #1004, opened 2026-09-20
Fix UBSan misaligned AggregateState access in hash aggregate scan

adsharma:
Fixes the ASan+UBSan nightly failure (`map_aggregate.cpp:106: member access within misaligned address ... for type 'struct AggregateState'`).

**Root cause:** factorized-table tuples are packed without alignment padding, so an aggregate state column can start at an odd offset (e.g. after a small group-by key). `HashAggregateScan` reinterpreted those bytes as `AggregateState*` and invoked virtuals on the misaligned pointer:
- `writeAggResultWithoutNullToVector` (`aggregateState->writeToVector()`) loads the vptr from a misaligned address; the WithNull path has the same problem via `constCast<AggregateStateWithNull>().isNull`.
- `hash_aggregate_scan.cpp`: `offset += aggState->getStateSize()` is a virtual call on the same misaligned pointer.

**Fix** (follows the memcpy precedent of fafc36265/2092c4284):
- `map_aggregate.cpp`: capture each aggregate's state size at plan time and copy the state bytes into an aligned thread-local buffer before any virtual dispatch in the move-to-vector funcs. The copy carries the vptr bit pattern, so the dynamic type is preserved; thread-local keeps it race-free across scan threads.
- `hash_aggregate_scan.cpp`: advance the scan offset using the table schema's state column sizes (sized with `AggregateFunction::getAggregateStateSize()` at plan time) instead of the virtual `getStateSize()`.

Nightly run: https://github.com/LadybugDB/ladybug/actions/runs/35438024832/job/105947162035

Note: that log contains further reports of the same *class* of bug in other aggregate states (`sum.h`, `min_max.h`, `collect.cpp`, `histogram.cpp`, `percentile_*.cpp` update/combine paths, `base_hash_table.cpp` `list_t`). Left for follow-up PRs to keep this one focused.


adsharma:
Addressed via #1014

